PR #5, opened 2019-06-07
CMakeLists.txt changes

SuperWig:
Bug Fixes:
* Currently can't use `find_package` as CMake is looking for `Config`not `Targets`
* Currently `magic_enum.hpp` is being installed to `include/include/magic_enum.hpp`

Additions/Removal:
* Added an alias to support using magic_enum via `add_subdirectory`
* Added `cxx_std_17` compile feature which requires 3.8 hence the change there
* Removed `GNUInstallDirs` since it wasn't being used
* Removed `PUBLIC_HEADER` and `INCLUDES DESTINATION` from the target install command as they are redundant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿cmake_minimum_required(VERSION 3.6)
+﻿cmake_minimum_required(VERSION 3.8)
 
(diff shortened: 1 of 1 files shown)


Neargye:
Thank! Сhanges look very useful.

PR #7, opened 2019-07-13
Fixing Syntax Error caused by "Windows.h"

GameForPeople:
While I was using "magic_enum" in my project, a syntax error occurred and I checked it, and I fixed it.
(Windows.h define macro max, min...wow;;)

I use "magic_enum" well. I respect your source code and appreciate it sincerely.
Thank you very much. :) 

 - Using Google Translation


Neargye:
Thank you very much for the corrections. I always forget the max/min macros in windows.h

PR #13, opened 2019-09-25
ensure O(1) enum value lookup

Trass3r:
Most compilers can't optimize away the `names` stack array in `name(E value)`: https://godbolt.org/z/ySzprC
Thoughts?


Neargye:
wip https://github.com/Neargye/magic_enum/issues/14

PR #16, opened 2019-10-03
Fix regression of enum_cast<>() when used with refs (EnumT& vs EnumT)

alexkaratarakis:
Watered-down/Min-repro of real use-case:
```C++
enum class MyEnum
{
    A,
    B,
    C,
};

int main()
{
    MyEnum a = MyEnum::A;
    auto& type = a;
    auto t = magic_enum::enum_cast<decltype(type)>(0);

    assert(t == MyEnum::A);

    return 0;
}
```

This used to work with 0.5, but doesn't work with 0.6.1.

- Despite the repetition, did not specify `D = std::decay_t<E>` as a template parameter because this opens the possibility for users manually specifying the second parameter.
- Added a regression test for it. Not sure if it is too obscure?

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -437,7 +437,7 @@ template <typename E>
 // Obtains enum value from integer value.
 // Returns std::optional with enum value.
 template <typename E>
-[[nodiscard]] constexpr auto enum_cast(std::underlying_type_t<E> value) noexcept -> detail::enable_if_enum_t<E, std::optional<std::decay_t<E>>> {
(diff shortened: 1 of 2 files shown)


Neargye:
Oh, Thank you for this case. 
I will expand the tests further, and check where it is also need decay enum type.

Neargye:
I tried to solve the сv-ref problem for all magic_enum API https://github.com/Neargye/magic_enum/commit/0b262f7cc89914b79a8b7840c89db4a2c850909a

Neargye:
FYI, release v0.6.2 with fixes https://github.com/Neargye/magic_enum/releases/tag/v0.6.2.
If you use vcpkg, I creat pr to update https://github.com/microsoft/vcpkg/pull/8500.

PR #18, opened 2019-10-11
Introduce mixed_sign_less()/min()/max()

alexkaratarakis:
Previously, there was mixed sign comparison:
```
reflected_min_v = -120 > 0U ? -120 : 0
```
which returns -120 because the condition is true, since the
compiler casts the signed value to unsigned, leading to a huge value.
This caused the following values as result:

```
reflected_min_v == -120
min_v = -1
static_cast<U>(min_v) = 18446744073709551615
```

The last one is used in magic_enum::enum_index().

The huge value was caught by static analysis (coverity), so this might get flagged by others.
This issue was introduced [here](https://github.com/Neargye/magic_enum/commit/370a1d523684532eb743ec78213fd323f4b011f3)

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -141,6 +141,34 @@ constexpr std::string_view pretty_name(std::string_view name) noexcept {
   return {}; // Invalid name.
 }
 
+template<typename L, typename R>
+constexpr bool mixed_sign_less(L left, R right) noexcept {
+  static_assert(std::is_integral_v<L>, "L must be an integral value");
+  static_assert(std::is_integral_v<R>, "R must be an integral value");
+  if constexpr(std::is_signed_v<L> == std::is_signed_v<R>) {
+    // If same signedness (both signed or both unsigned)
+    return left < right;
+  }
+  else if constexpr(std::is_signed_v<L>) {
+    // if 'left' is negative, then result is 'true', otherwise cast & compare
+    return left < 0 || static_cast<std::make_unsigned_t<L>>(left) < right;
+  }
+  else { // std::is_signed_v<R>
+    // if 'right' is negative, then result is 'false', otherwise cast & compare
+    return right >= 0 && left < static_cast<std::make_unsigned_t<R>>(right);
+  }
+}
+
+template<typename L, typename R>
+constexpr int mixed_sign_min_as_int(L left, R right) noexcept {
+  return mixed_sign_less(left, right) ? static_cast<int>(left) : static_cast<int>(right);
+}
+
+template<typename L, typename R>
+constexpr int mixed_sign_max_as_int(L left, R right) noexcept {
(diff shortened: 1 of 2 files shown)


include/magic_enum.hpp
```diff
@@ -141,6 +141,34 @@ constexpr std::string_view pretty_name(std::string_view name) noexcept {
...
+constexpr bool mixed_sign_less(L left, R right) noexcept {
+  static_assert(std::is_integral_v<L>, "L must be an integral value");
+  static_assert(std::is_integral_v<R>, "R must be an integral value");
+  if constexpr(std::is_signed_v<L> == std::is_signed_v<R>) {
+    // If same signedness (both signed or both unsigned)
+    return left < right;
+  }
+  else if constexpr(std::is_signed_v<L>) {
+    // if 'left' is negative, then result is 'true', otherwise cast & compare
+    return left < 0 || static_cast<R>(left) < right;
```
Neargye: How about `return left < 0 || static_cast<std::make_unsigned_t<L>>(left) < right;`
For fix it L::max() > R::max()

include/magic_enum.hpp
```diff
@@ -141,6 +141,34 @@ constexpr std::string_view pretty_name(std::string_view name) noexcept {
...
+  }
+  else { // std::is_signed_v<R>
+    // if 'right' is negative, then result is 'false', otherwise cast & compare
+    return right >= 0 && left < static_cast<L>(right);
```
Neargye: Same `return right >= 0 && left < static_cast<std::make_unsigned_t<R>>(right);`

Neargye:
This is a really good catch.

Neargye:
How about return left < 0 || static_cast<std::make_unsigned_t<L>>(left) < right;
For fix it L::max() > R::max()

alexkaratarakis:
Good point, there might be narrowing with the current `static_cast<R>`. I will test with `std::make_unsigned`.

alexkaratarakis:
Indeed there was a bug with `L::max() > R::max()`.
Fixed and added tests for this.

These were failing before:
```C++
REQUIRE(!mixed_sign_less(std::numeric_limits<int64_t>::max() - 17, std::numeric_limits<uint32_t>::max()));
REQUIRE(mixed_sign_less(std::numeric_limits<uint32_t>::max(), std::numeric_limits<int64_t>::max()));
REQUIRE(mixed_sign_less(std::numeric_limits<uint32_t>::max(), std::numeric_limits<int64_t>::max() - 17));
```

Interestingly, this was succeeding:
```C++
REQUIRE(!mixed_sign_less(std::numeric_limits<int64_t>::max(), std::numeric_limits<uint32_t>::max()));
```
because:
```C++
static_cast<uint32_t>(std::numeric_limits<int64_t>::max()) == std::numeric_limits<uint32_t>::max()
```
This corner case is why I added the '17' offset.

Neargye (approved):

PR #19, opened 2019-10-11
Change the way range_min()/max() is calculated

alexkaratarakis:
This was previously done via parameter expansion. The expanded form
looks like this:
A(128) && A(127) && ... && A(X) && ... && A(0)

where A(X) is the actual range_max which is crafted to return 'false'.
Static analysis detects that expressions after A(X) are
dead code due to short-circuiting.

The new implementation uses a simple loop to find the extrema,
similarly to the loop in detail::values().
Furthermore, it uses only one parameter expansion (for reflected_set_v),
as opposed to two (one for range_min and one for range_max).

Notes:
I investigated this because it was flagged by Coverity static analysis. While technically not incorrect, I think this flagging is overzealous and will report to Coverity because it looks like all [fold expressions](https://en.cppreference.com/w/cpp/language/fold) involving `&&` and `||` will be getting flagged with 'dead code' issues.

However, I think the proposed implementation has merits on its own right, so it might be worthwhile to apply.

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -204,29 +204,49 @@ constexpr std::size_t reflected_size() {
   return static_cast<std::size_t>(size);
 }
 
+template <typename E>
+inline constexpr std::size_t reflected_size_v = reflected_size<E>();
+
 template <typename E, int... I>
-constexpr int range_min(std::integer_sequence<int, I...>) noexcept {
+constexpr std::array<bool, reflected_size_v<E>> reflected_set_v_helper(std::integer_sequence<int, I...>) {
+  return {{(n<E, static_cast<E>(I + reflected_min_v<E>)>().size() != 0)...}};
+}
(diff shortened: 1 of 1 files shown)


alexkaratarakis:
Investigating the Windows failures.

edit: fixed!

Neargye:
Yes, the new code does the same, and in general it is possible for the compiler to understand the new code even easier.

Neargye:
I will be happy to merge your pr next week, now I speak at the conference, so there is no many time at the computer.

Neargye (approved):

Neargye:
This pr made me think about calculating values, so I reorganized values/min/max calculated in https://github.com/Neargye/magic_enum/commit/4066fa35e5bb3bc30b6825741ae2b208e6726202.
I tried to remove unnecessary computations, which on tests positively affected compilation time.

PR #24, opened 2020-01-31
improve compiler version check

pck:
Hello

The compiler version check shouldn't use string comparison.
This causes problems with clang 10 and 11.

```
$ conan create . pck/stable -s compiler=clang -s compiler.version=10 -s compiler.libcxx=libc++
...
ERROR: magic_enum/0.6.4@pck/stable: Invalid configuration: magic_enum: Unsupported compiler (https://github.com/Neargye/magic_enum#compiler-compatibility).
```

Thank you for sharing your work.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from conans import ConanFile
+from conans import ConanFile, tools
 from conans.errors import ConanInvalidConfiguration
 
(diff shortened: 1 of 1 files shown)


Neargye:
Hi @pck,

Thank you for your improvements. I'm not very good at setting up Conan, so I'm happy with these improvements.

Neargye:
@pck upload new v0.6.5 with this fix on Conan.

PR #30, opened 2020-03-13
 Introduce contains_value()

oficsu:
**Why**? This is useful for validation. In some cases, we receive some numeric value from outside our program (it may be some serialization library or REST API), this often requires validation of value. 

**Why not use enum_index()**? There are two reasons. 

First, the name of this function has little different meaning and the code below is a little confused:
```cpp
auto has_value = enum_index(enum_value).has_value();
```
I think, this code is more readable:
```cpp
auto has_value = contains_value<E>(value);
```
The second reason is that we need an enum instance, but in these cases, we usually have a raw data or an integral value, so we need to use static_cast<E>(value), but this makes user code harder to read and requires an explicitly defined underlying type because otherwise cast is causes undefined behavior

**About naming**. I'm not sure if this is the best name for this function. It can be renamed to has_value() or enum_has_value() (since this library prefers naming starting with enum_ )

**About implementation**. I changed enum_traits::index() to avoid undefined behavior described above, so now there are two overloads of this function. The first accepts value of the underlying type and is implemented as before.
The second just casts enumerator to the underlying type (casting to the underlying type from enum is safe unlike casting to enum from the underlying type) and calls first overloading.
Another nice change is that there is no more need for many casts to underlying type.
Maybe it wasn’t worth do two overloads in enum_traits, but I don’t know how to do it better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Header-only C++17 library provides static reflection for enums, work with any en
 * `enum_names` obtains string enum name sequence.
 * `enum_entries` obtains pair (value enum, string enum name) sequence.
 * `enum_index` obtains index in enum value sequence from enum value.
+* `contains_value` checks whether enum contains such value
 * `is_unscoped_enum` checks whether type is an [Unscoped enumeration](https://en.cppreference.com/w/cpp/language/enum#Unscoped_enumeration).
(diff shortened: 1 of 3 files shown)


Neargye:
Hi @oficsu,

I really like changes in enum_traits

I would suggest two things:
* `int index(underlying_type value)` make private
* About naming `enum_contains` and two overloads for enum and underlying_type.

PR #31, opened 2020-03-13
Change contains_value to enum_contains

Neargye:


Neargye:
@oficsu , could you check this changes?

> Values of integer, floating-point, and enumeration types can be converted by static_cast or explicit cast, to any enumeration type. If the underlying type is not fixed and the source value is out of range, the result is unspecified (until C++17)undefined (since C++17). (The source value, as converted to the enumeration's underlying type if floating-point, is in range if it would fit in the smallest bit field large enough to hold all enumerators of the target enumeration.) Otherwise, the result is the same as the result of implicit conversion to the underlying type.

In this case, cast underlying_type to enum seems ok.

oficsu:
Yes, I like these changes.

oficsu (approved):

PR #56, opened 2020-10-02
Add support for bazel

FelixDuvalletKodiak:
Bazel support for the magic_enum repository would help others using
bazel integrate this library into their projects.

This PR adds a BUILD file which enables magic_enum to be imported
seamlessly into other WORKSPACE files without having to write a custom
BUILD file.

It also adds bazel binaries & tests to enable running them via bazel
locally:

```
bazel build //... 
bazel test //... 
bazel run //:example
```

Finally, bazel compilation artifacts are added to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
 _deps
+
+### Bazel build artifacts ###
+bazel-*
--- a/.gitignore
+++ b/.gitignore
@@ -50,4 +50,4 @@ CTestTestfile.cmake
 _deps
 
 ### Bazel build artifacts ###
-bazel-*
+/bazel-*
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,56 @@
+licenses(["notice"])
+
+exports_files(["LICENSE"])
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "magic_enum",
+    hdrs = ["include/magic_enum.hpp"],
+    strip_include_prefix = "include",
+)
+
+cc_binary(
+    name = "example",
+    srcs = ["example/example.cpp"],
+    deps = [":magic_enum"],
+    copts = ["-std=c++17"],
+)
+
+cc_binary(
+    name = "enum_flag_example",
+    srcs = ["example/enum_flag_example.cpp"],
+    deps = [":magic_enum"],
+    copts = ["-std=c++17"],
+)
+
+cc_binary(
+    name = "example_custom_name",
+    srcs = ["example/example_custom_name.cpp"],
+    deps = [":magic_enum"],
+    copts = ["-std=c++17"],
+)
+
+cc_library(
+    name = "catch",
+    srcs = [],
+    hdrs = ["test/3rdparty/Catch2/catch.hpp"],
+    strip_include_prefix = "test/3rdparty/Catch2",
+)
+
+cc_test(
+    name = "test",
+    srcs = [
+        "test/test.cpp",
+    ],
+    deps = [":magic_enum", ":catch"],
+    copts = ["-std=c++17"],
+)
+cc_test(
+    name = "test_flags",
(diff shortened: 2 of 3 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,56 @@
+licenses(["notice"])
+
+exports_files(["LICENSE"])
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "magic_enum",
+    hdrs = ["include/magic_enum.hpp"],
+    strip_include_prefix = "include",
```
HackAttack: Have you tested actually consuming this as an external dependency? I ask because IIRC `strip_include_prefix` doesn’t work in that case because there will be an extra `external/` to contend with. I have not tested that your change _doesn’t_ work, but the below modification definitely does.

```suggestion
    includes = ["include"],
```

.gitignore
```diff
@@ -48,3 +48,6 @@ install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
 _deps
+
+### Bazel build artifacts ###
+bazel-*
```
HackAttack: [Nit] This will ignore `bazel-*` in all directories, but you actually only care about the symlinks created at the root of the repo.

```suggestion
/bazel-*
```

BUILD.bazel
```diff
@@ -0,0 +1,56 @@
+licenses(["notice"])
+
+exports_files(["LICENSE"])
```
HackAttack, line 3: Is this needed?

README.md
```diff
@@ -219,6 +219,17 @@ CPMAddPackage(
 )
 ```
 
+Bazel is also supported, simply add to your WORKSPACE file:
+```
+git_repository(
```
HackAttack: [`http_archive` is preferred](https://docs.bazel.build/versions/3.5.1/external.html#repository-rules) so I would use that as the example.

README.md
```diff
@@ -219,6 +219,17 @@ CPMAddPackage(
...
+    name = "magic_enum",
+    commit = <>,
+    remote = "https://github.com/Neargye/magic_enum",
+)
+```
+To use bazel inside the repository it's possible to do:
+`export CC=clang ; bazel build //... ; bazel test //... ; bazel run //:example`
```
HackAttack: Specifying `clang` is confusing, since right below it says that multiple compilers are supported.

README.md
FelixDuvalletKodiak: Thanks for the suggestion, done.

README.md
FelixDuvalletKodiak: I updated the comment, please let me know what you think.
I initially ran into an issue because the system-default compiler is not supported, but you're right that any supported can be used.

BUILD.bazel
FelixDuvalletKodiak: AFAIK bazel only has _plans_ to support license files as part of the build, but since the repo provides it I am following (what I believe to be) an accepted convention of exporting it.
GoogleTest does this, [for instance](https://github.com/google/googletest/blob/master/BUILD.bazel).

.gitignore
FelixDuvalletKodiak: Good point, thanks.

BUILD.bazel
FelixDuvalletKodiak: Yes, that's much nicer. Thanks for the suggestion, done.

FelixDuvalletKodiak:
@HackAttack thanks for the review, please take another look when you get a chance.

Neargye:
@FelixDuvalletKodiak Thank for your pr.

Let's wait for a review from @HackAttack , because I myself do not use bazel.

FelixDuvalletKodiak:
@HackAttack friendly ping, are there further improvements we can make to this PR before merging it?

HackAttack (approved):

Neargye:
@FelixDuvalletKodiak @HackAttack thanks!

Neargye (approved):

PR #70, opened 2021-01-08
Detect values out of range and fail compilation

bjacklyn:
This checks for a value at (range_min - 1) and (range_max + 1) and fails compilation with a static_assert if any value is found.

I put the check in `values()` -- is that the only location needed to enable it globally?

In my org it is preferable for this to be caught at build (compile) time rather than at runtime. Usually what happens is someone merges a PR that adds one more entry to an enum which exceeds the range_max, and it isn't detected until the software is behaving weirdly at runtime. We are using the range 0 - 64, and then we have to specialize for anything larger. `magic_enum` is unfortunately quite slow during compilation so narrowing the default range this much saved ~10% overall build time, but at the cost of needing to specialize (and running into this issue) more often.

Regarding implementation, technically this might break anyone who actually wants magic_enum to be missing some values, so maybe it should be behind some `#if` check that is enabled like `-DENABLE_MAGIC_ENUM_OFF_BY_ONE_RANGE_CHECK`. Also maybe this could be cleaner if refactored so that `min` and `max` were in `values()` and there was separate functions for checking min/max that could be each called by 2 `static_asserts` in `values()`.

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -435,11 +435,30 @@ constexpr auto values(std::index_sequence<I...>) noexcept {
 }
 
 template <typename E, bool IsFlags, typename U = std::underlying_type_t<E>>
+inline static constexpr void detect_values_out_of_range()
+{
+  constexpr auto min = reflected_min_v<E, IsFlags>;
+  constexpr auto max = reflected_max_v<E, IsFlags>;
+  constexpr auto range_size = max - min + 1;
+
+  if constexpr (cmp_less((std::numeric_limits<U>::min)(), min) && !IsFlags)
+  {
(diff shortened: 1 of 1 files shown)


Neargye:
Thanks for pr. Yes, check this on function values() is ok.

Neargye:
I would rather not do a separate function, just check in values()

Trass3r:
> We are using the range 0 - 64, and then we have to specialize for anything larger. `magic_enum` is unfortunately quite slow during compilation so narrowing the default range this much saved ~10% overall build time

Unfortunately clang's `-ftime-trace` doesn't help much with constexpr functions but maybe it could still help a bit.

PR #75, opened 2021-02-22
chore: update catch2 to 2.13.4 for Apple Silicon support

gjasny:
Additionally I included the Catch2 header with `catch2/catch.hpp` because that's how upstream installs and uses it nowadays.


Neargye:
Why two folders now Сatch2/catch2?

BUILD.bazel
```diff
@@ -34,7 +34,7 @@ cc_binary(
 cc_library(
     name = "catch",
     srcs = [],
-    hdrs = ["test/3rdparty/Catch2/catch.hpp"],
+    hdrs = ["test/3rdparty/Catch2/catch2/catch.hpp"],
```
Neargye: Why folders now Сatch2/catch2?

gjasny: I chose that structure to isolate the 3rdpartly libs from each other. Right now the include path for catch is `test/3rdparty/Catch2`, if I omit the "extra" level it would be `test/3rdparty` which feels a little bit too broad.

If you still want me to change it, just drop me a comment.

Neargye: maybe then `test/3rdparty/Catch2/include/catch2/catch.hpp`?

Neargye:
Will wait for CI and then merge

Neargye (approved):

PR #84, opened 2021-05-26
Add ARCH_INDEPENDENT so the generated cmake config files are really arch independent

### LINKED ISSUE #83: cmake "bug" about header only library
Hi,

in the code lines

https://github.com/Neargye/magic_enum/blob/8deb7f96e74e62cd1d9fd8920debfbe27f4bb8d1/CMakeLists.txt#L33-L35

the package config is generated.

I thought magic_enum is a header only library but the package config file generated in those lines still has an architecture check active, which will cause

```
The following configuration files were considered but not accepted:
1> [CMake] 
1> [CMake]     C:/Code/magic_enum/install/lib/cmake/magic_enum/magic_enumConfig.cmake, version: 0.7.2 (64bit)
```
when trying to use the same install for a 32bit build.

only adding `ARCH_INDEPENDENT` to the `write_basic_package_version_file` disables this check.

Flow86:
downside is that it at least requires cmake 3.14

closes #83

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿cmake_minimum_required(VERSION 3.8)
+﻿cmake_minimum_required(VERSION 3.14)
 
 project(magic_enum VERSION "0.7.2" LANGUAGES CXX)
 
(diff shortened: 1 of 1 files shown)


Neargye:
https://github.com/xtensor-stack/xtl/issues/225 maybe add fallback for CMake <3.14 ?

Neargye:
Or does it no longer make sense to support the сmake below 3.14?

Flow86:
> Or does it no longer make sense to support the сmake below 3.14?

even debian has a current 3.16 available? but I can't decide that ;-)

Neargye (approved):

PR #86, opened 2021-06-17
Add ability to include external configuration header

sajimon:
Hello, 
I propose to add ability to include external user configuration file that will define all required constants, instead of using includes order (although this option stays available as well). 

MOTIVATION:
I'm working within tight embedded environment and I'm using ETL library instead of STL - it is more lightweight and avoids dynamic mem. alloc. 
While using documented method of setting for example MAGIC_ENUM_USING_ALIAS_STRING (by proper including order) is functional is also prone to errors and if other member of the team accidentally include magic_enum.hpp directly, std::string may be pulled. 
Approach with external cfg. file makes it far less likely.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -38,6 +38,22 @@
   #define MAGIC_ENUM_USING_ALIAS_OPTIONAL template <typename T> using optional = my_lib::Optional<T>;
   #include <magic_enum.hpp>
   ```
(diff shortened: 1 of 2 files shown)


Neargye:
@sajimon could you please add some info in the doc about the config header?

Neargye:
Thanks!

PR #88, opened 2021-06-25
Fix underflow in out of range check

alexkaratarakis:
The previous fix addressed bool but it seems that was not the heart
of the issue.

The problem is that (lhs - 1) can underflow.

Another example of the check trigger when it shouldn't is:
uint8_t with min_value set to 0 and a value of uint8_t::max
lhs-1 would underflow and find max (see added test).

The issue happens when lhs == rhs. To avoid that situation: flip the if-else!
`cmp_less(rhs, lhs)`means that rhs is strictly less than lhs, so
the check will not trigger when lhs == rhs.

Also, cleanup the bool special-case, as it is no longer necessary.

Note: for the bug to manifest, it is important that the min
of the enums is customized to be 0, so corresponding enum_range
specializations need to be present in the regression tests.

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -276,10 +276,13 @@ constexpr bool cmp_equal(string_view lhs, string_view rhs, BinaryPredicate&& p) 
 template <typename L, typename R>
 constexpr bool cmp_less(L lhs, R rhs) noexcept {
   static_assert(std::is_integral_v<L> && std::is_integral_v<R>, "magic_enum::detail::cmp_less requires integral type.");
-
   if constexpr (std::is_signed_v<L> == std::is_signed_v<R>) {
     // If same signedness (both signed or both unsigned).
     return lhs < rhs;
+  } else if constexpr (std::is_same_v<L, bool>) { // bool special case due to msvc's C4804, C4018
+      return static_cast<R>(lhs) < rhs;
+  } else if constexpr (std::is_same_v<R, bool>) { // bool special case due to msvc's C4804, C4018
+      return lhs < static_cast<L>(rhs);
   } else if constexpr (std::is_signed_v<R>) {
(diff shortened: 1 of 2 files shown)


include/magic_enum.hpp
```diff
@@ -276,10 +276,13 @@ constexpr bool cmp_equal(string_view lhs, string_view rhs, BinaryPredicate&& p)
 template <typename L, typename R>
 constexpr bool cmp_less(L lhs, R rhs) noexcept {
   static_assert(std::is_integral_v<L> && std::is_integral_v<R>, "magic_enum::detail::cmp_less requires integral type.");
-
```
Neargye: Could return a line break? purely for consistency

include/magic_enum.hpp
```diff
@@ -383,13 +386,11 @@ constexpr int reflected_min() noexcept {
...
     constexpr auto rhs = (std::numeric_limits<U>::min)();
 
-    if constexpr (std::is_same_v<bool, U>) {
-      return static_cast<int>(rhs);
-    } else if constexpr (cmp_less(lhs, rhs)) {
-      return rhs;
-    } else {
-      static_assert(!is_valid<E, value<E, lhs - 1, IsFlags>(0)>(), "magic_enum::enum_range detects enum value smaller than min range size.");
+    if constexpr (cmp_less(rhs, lhs)) {
+      static_assert(!is_valid<E, value<E, lhs-1, IsFlags>(0)>(), "magic_enum::enum_range detects enum value smaller than min range size.");
```
Neargye: `lhs-1` -> `lhs - 1`

Neargye:
Yes, the previous fix just handled the special case of the boolean as an enum type. and just did not trigger the comparison and check of the range

Neargye (approved):

Neargye:
Thanks for pr!

PR #90, opened 2021-07-03
Remove optional<> with std::decay (Mac OS / Xcode)

JasonCoombs:
I don't have a good explanation for why, but an optional<> type is considered by Clang on Mac OS (using Xcode) to be different from, and not convertible into, the same type without optional<> therefore magic_enum builds produced SFINAE errors, as the optional<> type does not appear to be an enum accordig to enable_if_enum_t<>. Solved by removing optional<> template instantiation around each std::decay<>

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -738,7 +738,7 @@ template <typename E>
 // Obtains enum value from integer value.
 // Returns optional with enum value.
 template <typename E>
-[[nodiscard]] constexpr auto enum_cast(underlying_type_t<E> value) noexcept -> detail::enable_if_enum_t<E, optional<std::decay_t<E>>> {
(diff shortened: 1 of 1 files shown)


JasonCoombs:
Maybe the decay<> could be used on the optional<> type, instead, to get the correct SFINAE behavior?

Neargye:
Could you show your compilation error?

Neargye:
I would also like to show an example of using the function
```cpp
  auto c3 = magic_enum::enum_cast<Color>(10);
  if (c3.has_value()) {
    std::cout << "GREEN = " << c3.value() << std::endl; // GREEN = 10
  }
```

[optional](https://en.cppreference.com/w/cpp/utility/optional) was added as a return type for functions that may not have a return value

Issue #94, opened 2021-08-05
Enum entries with non-ascii characters get removed


Emill:
```C++
enum Siffra {
    ETT,
    TVÅ,
    TRE
};
```

This code works fine in gcc and clang, but when using magic_enum with it, it only detects `ETT` and `TRE`.

It seems the `pretty_name` function in magic_enum only allows `0-9`, `a-z`, `A-Z` and `_`. Annex E in the C++ standard seems to allow these kind of characters in identifiers.

FYI, `__PRETTY_FUNCTION__` includes this for TVÅ: `[E = Siffra, V = TV\303\205]`.

Neargye:
Fixed #95

PR #95, opened 2021-08-08
Added support for non-ASCII characters (UNIX/Linux)

uruha-komachin:
Non-ASCII characters are allowed in C++ identifiers so I added support for them.

tbh, I have to admit that we barely see anyone write code with non-ASCII identifiers. Nevertheless, they're still legal C++ code and they compiles. IMO we should add support for this case.


uruha-komachin:
Also, this pr should solve #94 . Please advise if you think there are more necessary changes need to be made. Thank you 😃

Neargye:
@uruha-komachin could you please add some test too?

uruha-komachin:
Hi, I've added an option, `MAGIC_ENUM_OPT_ENABLE_NONASCII`, and some test cases for the non-ASCII support. 

However, as there is a [known bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67224) in GCC 9's UTF-8 support, we will skip the non-ASCII example and test for GCC version below 10.0.

As for Windows, the codepage thingy and unicode support on Windows is a PAIN. Therefore I didn't port this feature to Windows and decided to stop there. This feature will need some future help (of course, if anyone wants to implement this).

uruha-komachin:
Sorry, I saw the workflow failed, and it was me forgetting to skip the non-ASCII test in `test/CMakeLists.txt`. I've updated the file. It should be fine now.

Neargye:
@uruha-komachin I had the nerve to tweak the style a bit. If everything is ok, I'm ready to merge

uruha-komachin:
@Neargye Everything looks great to me. Please feel free to merge this pr 🎉

Neargye (approved):

PR #110, opened 2021-11-01
Some code quality

florianbecker:
* Remove redundant pairs of parentheses
* Remove inline from constexpr variable

What about making the defines constexpr?

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -132,10 +132,10 @@ struct enum_range {
 };
 
 static_assert(MAGIC_ENUM_RANGE_MIN <= 0, "MAGIC_ENUM_RANGE_MIN must be less or equals than 0.");
-static_assert(MAGIC_ENUM_RANGE_MIN > std::numeric_limits<std::int16_t>::min(), "MAGIC_ENUM_RANGE_MIN must be greater than INT16_MIN.");
+static_assert(MAGIC_ENUM_RANGE_MIN > (std::numeric_limits<std::int16_t>::min)(), "MAGIC_ENUM_RANGE_MIN must be greater than INT16_MIN.");
 
 static_assert(MAGIC_ENUM_RANGE_MAX > 0, "MAGIC_ENUM_RANGE_MAX must be greater than 0.");
-static_assert(MAGIC_ENUM_RANGE_MAX < std::numeric_limits<std::int16_t>::max(), "MAGIC_ENUM_RANGE_MAX must be less than INT16_MAX.");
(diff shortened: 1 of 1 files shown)


Neargye:
Please pay attention to this https://github.com/Neargye/magic_enum/pull/7, so please revert `(std::numeric_limits<std::int16_t>::min)()` and simular.

> making the defines constexpr

they are used to configure the library, read more [here](https://github.com/Neargye/magic_enum/blob/master/doc/limitations.md) and [here](https://github.com/Neargye/magic_enum/blob/master/doc/reference.md#synopsis)

Neargye:
> Remove inline from constexpr variable

This ok

include/magic_enum.hpp
```diff
@@ -126,16 +126,16 @@ namespace customize {
 template <typename E>
 struct enum_range {
   static_assert(std::is_enum_v<E>, "magic_enum::customize::enum_range requires enum type.");
-  inline static constexpr int min = MAGIC_ENUM_RANGE_MIN;
-  inline static constexpr int max = MAGIC_ENUM_RANGE_MAX;
+  static constexpr int min = MAGIC_ENUM_RANGE_MIN;
```
Neargye, line 129: This ok

florianbecker:
Mh, interesting. So maybe you could add this special case to the integration pipeline?

Ok, I'll revert that.

Is it in your interest to convert the defines to constexpr? So I could do that next.

Neargye:
>Is it in your interest to convert the defines to constexpr? So I could do that next.

Could you clarify what definitions are we talking about?

florianbecker:
Simple, replace these defines by constexpr or directly create the min/max constexpr.
#define MAGIC_ENUM_VERSION_MAJOR 0
#define MAGIC_ENUM_VERSION_MINOR 7
#define MAGIC_ENUM_VERSION_PATCH 3
#define MAGIC_ENUM_RANGE_MIN -128
#define MAGIC_ENUM_RANGE_MAX 128

not sure if this is a good idea - maybe, maybe not... 
#define MAGIC_ENUM_SUPPORTED 1
#define MAGIC_ENUM_SUPPORTED_ALIASES 1

Neargye:
> Is it in your interest to convert the defines to constexpr? So I could do that next.
No

All this macro need to configure or check. 
```cpp
#if MAGIC_ENUM_VERSION_MAJOR  = 0 && MAGIC_ENUM_VERSION_MINOR >= 7 // work
```
```cpp
constexpr auto MAGIC_ENUM_VERSION_MAJOR   = 7;
#if MAGIC_ENUM_VERSION_MINOR >= 7 // not work
```

florianbecker:
Ok, thank you. This is finished than 😊

Neargye:
@florianbecker  if you want, you can make similar changes for https://github.com/Neargye/nameof

PR #112, opened 2021-11-29
case insensitive enum_cast syntactic sugar

jiggersplat:
allows you to write magic_enum::enum_cast<foo>("bar", magic_enum::case_insensitive);

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -118,6 +118,7 @@ MAGIC_ENUM_USING_ALIAS_STRING
 using std::string;
 #endif
 
+ 
 namespace customize {
 
 // Enum value must be in range [MAGIC_ENUM_RANGE_MIN, MAGIC_ENUM_RANGE_MAX]. By default MAGIC_ENUM_RANGE_MIN = -128, MAGIC_ENUM_RANGE_MAX = 128.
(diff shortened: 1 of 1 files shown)


include/magic_enum.hpp
```diff
@@ -118,6 +118,7 @@ MAGIC_ENUM_USING_ALIAS_STRING
 using std::string;
 #endif
 
+
```
Neargye, line 121: remove this

include/magic_enum.hpp
```diff
@@ -746,6 +747,11 @@ template <typename E>
   return detail::entries_v<D>;
 }
 
+// allows you to write magic_enum::enum_cast<foo>("bar", magic_enum::case_insensitive);
+auto case_insensitive = [](char lhs, char rhs) {
```
Neargye, line 751:
```cpp
inline constexpr auto case_insensitive = [](char lhs, char rhs) noexcept constexpr -> bool {
#if defined(MAGIC_ENUM_ENABLE_NONASCII)
  static_assert(sizeof(decltype(this) == 0), "magic_enum::case_insensitive not supported Non-ASCII feature.");
#endif
  return detail::to_lower(lhs) == detail::to_lower(rhs);
};
```

Neargye:
please add a test.

Neargye:
Thanks! will be done in #139

PR #120, opened 2022-01-05
New enum-flags API

### LINKED ISSUE #119: Combining flags ui and non-flags ui in the same processing routine
I am running a project that processes many enums to load them as read-only properties into a Lua context using [`LuaBridge`](https://github.com/vinniefalco/LuaBridge). With your help I was able to get the `flags` ui working for my power-of-2 enums, and it greatly reduces the compile time and memory requirements for compiling them.

Now I would like to use a single routine to process either type of enum. So far, this is working, where `Process` is a runtime function that adds the property to `LuaBridge`:

```c++
template<typename ENUM, bool IsFlags = false>
void AddEnumeratedType(const std::string& nameSpace, lua_State *l)
{
   constexpr std::size_t count = []()
            {
               if constexpr (IsFlags)
                  return magic_enum::flags::enum_count<ENUM>();
               else
                  return magic_enum::enum_count<ENUM>();
            }();
   constexpr auto names = []()
            {
               if constexpr (IsFlags)
                  return magic_enum::flags::enum_names<ENUM>();
               else
                  return magic_enum::enum_names<ENUM>();
            }();
   constexpr_for<0, count, 1>([&](auto i)
   {
      if constexpr (IsFlags)
         Process(names[i].data(), magic_enum::flags::enum_value<ENUM, i>());
      else
         Process(names[i].data(), magic_enum::enum_value<ENUM, i>());
   });
}
```

But it's a little ugly. The main requirement is that only the `flags` or the non-`flags` path should be evaluated. Not both. It seems like if there were a combined ui that took both `ENUM` and `IsFlags` as template parameters then the code would be a lot nicer. (Many of your internal routines do this.)

(hypothetical code with `combined` namespace, but just for clarity in this example)

```c++
template<typename ENUM, bool IsFlags = false>
void AddEnumeratedType(const std::string& nameSpace, lua_State *l)
{
   constexpr std::size_t count = magic_enum::combined::enum_count<ENUM, IsFlags>();
   constexpr auto names = magic_enum::combined::enum_names<ENUM, IsFlags>();

   constexpr_for<0, count, 1>([&](auto i)
   {
      Process(names[i].data(), magic_enum::combined::enum_value<ENUM, IsFlags, i>());
   });
}
```

The version I have now is working, so this is definitely just a nice-to-have kind of suggestion. Also, perhaps there is a cleaner way to do it that I haven't thought of, using the existing code in `magic_enum`. Either way, it's a fantastic tool for reflecting enums.

Neargye:
Fix #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,8 +197,6 @@ enum class Color { RED = 2, BLUE = 4, GREEN = 8 };
 
 * Before use, read the [limitations](doc/limitations.md) of functionality.
 
-* For the small enum use the API from the namespace `magic_enum`, and for enum-flags use the API from the namespace `magic_enum::flags`.
-
 ## Integration
 
 * You should add the required file [magic_enum.hpp](include/magic_enum.hpp).
--- a/doc/limitations.md
+++ b/doc/limitations.md
@@ -28,14 +28,12 @@
 
     enum class number { one = 100, two = 200, three = 300 };
 
-    namespace magic_enum::customize {
     template <>
-    struct enum_range<number> {
+    struct magic_enum::customize::enum_range<number> {
       static constexpr int min = 100;
       static constexpr int max = 300;
       // (max - min) must be less than UINT16_MIN.
     };
-    } // namespace magic_enum
     ```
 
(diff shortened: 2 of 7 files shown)


Neargye:
TODO:
- [x] Add enum_flags_name to doc
- [x] Add flags settings to doc
- [x] is_sparse for flags

PR #123, opened 2022-01-25
Rgp static assert for static string

### LINKED ISSUE #122: Visual Studio 2022 IntelliSense Error E0095
I am encountering a strange IntelliSense error in Visual Studio 2022. This is not actually causing the compile to fail or any other issues that I can see. And even weirder, when I click on the error to view the line of code, the error disappears. (I haven't figured out what makes it reappear.) The line of code is line 183:

```c++
  char chars_[N + 1];
```

The error message is here:

Screen Shot 2022-01-24 at 11 19 14 AM

For all I know it is a bug in VS 2022. But I thought I would report it in case it is something you wish to know about.

rpatters1:
This pull request is to resolve #122.


Neargye:
Thanks!

PR #127, opened 2022-02-05
Add enum fusing function

pavelkryukov:
Emulates 2D switch/case statements

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Header-only C++17 library provides static reflection for enums, work with any en
 * `enum_index` obtains index in enum value sequence from enum value.
 * `enum_contains` checks whether enum contains enumerator with such value.
 * `enum_type_name` returns name of enum type.
+* `enum_fuse` allows multidimensional switch/cases.
 * `is_unscoped_enum` checks whether type is an [Unscoped enumeration](https://en.cppreference.com/w/cpp/language/enum#Unscoped_enumeration).
 * `is_scoped_enum` checks whether type is an [Scoped enumeration](https://en.cppreference.com/w/cpp/language/enum#Scoped_enumerations).
 * `underlying_type` improved UB-free "SFINAE-friendly" [underlying_type](https://en.cppreference.com/w/cpp/types/underlying_type).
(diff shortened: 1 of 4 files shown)


Neargye:
`enum_index(enums).value()` can throw an exception, is it possible to somehow handle situations when there are no values? return 0?

pavelkryukov:
For me, throwing an exception sounds right:
- in runtime (`switch`) context it captures invalid input
- in constexpr (`case`) context it prevents compilation.

If we pass 0 to the pairing function for "no-value" case (or return "0" overall):
- compilation will succseed, but `case` will be silently unreachable
- if one "no-value" is passed to `switch` and `case` handles another, false match will appear
- if there are two different "no-value" cases, compiler will raise unclear "duplicate case" error

Examples:
```c++
switch (magic_enum::enum_fuse(Color::RED, Color(16)))
{
case magic_enum::enum_fuse(Color::RED, Color(17)): // also matched
case magic_enum::enum_fuse(Color::RED, Color(16)): // duplicate case error
}
```

Do you know how to break constexpr compilation but allow returning 0 in runtime?

pavelkryukov:
Also, current implementation with Cantor pairing is the safest, but slow.
The faster apporach is `or`-ing shifted first enum with the second one.
It would require stronger guarantees to perform correct shift.

Neargye:
> Do you know how to break constexpr compilation but allow returning 0 in runtime?

Only in c++20 with https://en.cppreference.com/w/cpp/types/is_constant_evaluated

```cpp
if (std::is_constant_evaluated() && not_valid()) {
throw ...
}
```

pavelkryukov:
I tried this:

```c++
template<typename ... Es>
[[nodiscard]] constexpr size_t enum_fuse(Es ... enums) noexcept
{   
  static_assert(sizeof...(Es) >= 2, "magic_enum::enum_fuse requires at least 2 enums");
  if (!(enum_index(enums).has_value() && ...)) {
#ifdef __cpp_lib_is_constant_evaluated      
    if (std::is_constant_evaluated())
      throw std::logic_error("magic_enum::enum_fuse accepts only in-range enum values"); 
#endif
    return 0;
  }

  // Add 1 to prevent matching 2D fusions with 3D fusions etc.
  return detail::cantor_pair((enum_index(enums).value() + 1)...);
}
```

but GCC 10 fairly complains I throw an exception in a `noexcept` function.

Neargye:
Thanks.

Neargye:
About `noexcept ` in this case is not needed, because `value()` can throw exceptions, and if the function is noexcept, then the std::terminate() will be called.

pavelkryukov:
> because value() can throw exceptions, and if the function is noexcept, then the std::terminate() will be called.

I was under impression you requested to make the function `return 0` keeping `noexcept` and avoiding termination.
So now I'm puzzled that final version throws custom exception only in constant evaluated context, and runtime throw default one, which is far less informative.

Neargye:
fixed

PR #137, opened 2022-02-18
support gcc8 with defined MAGIC_ENUM_NO_CHECK_SUPPORT and with specialized magic_enum::customize::enum_name

schaumb:
I tried to run examples and tests with gcc8, with manually specialized `magic_enum::customize::enum_name`.
I know that GCC8 is not supported, but these changes help us to make compatible our code with older GCC too only with the custom template function.

---

I replaced 
`#if defined(MAGIC_ENUM_SUPPORTED) && MAGIC_ENUM_SUPPORTED`
macros to 
`if constexpr (supported<E>::value) {`

at class name getter `n()`, and  `operator<<`
These functions work with custom `magic_enum::customize::enum_name`, so it is enough to check `supported` flag.

---

I replaced  `detail::min_v<E>` template parameter `E`, because `E` can be a reference, and `std::underlying_type` is undefined in some compilers when the template parameter is a reference.

---

I switched `values_v` and `entries_v` from `auto` to `std::array`, because GCC8 is not able to deduct the whole type, but `std::array` template parameters. This change should not change the newer compilers (and it is not a big change)

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -336,11 +336,11 @@ constexpr auto n() noexcept {
 #  elif defined(_MSC_VER)
       constexpr auto name = pretty_name({__FUNCSIG__, sizeof(__FUNCSIG__) - 17});
 #  else
-      constexpr std::string_view name = "";
+      constexpr string_view name = "";
 #  endif
       return static_string<name.size()>{name};
   } else {
-      return std::string_view{}; // Unsupported compiler.
+      return string_view{}; // Unsupported compiler.
   }
 }
 
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -534,7 +534,7 @@ constexpr auto names(std::index_sequence<I...>) noexcept {
 }
 
 template <typename E>
-inline constexpr auto names_v = names<E>(std::make_index_sequence<count_v<E>>{});
+inline constexpr std::array names_v = names<E>(std::make_index_sequence<count_v<E>>{});
 
 template <typename E, typename D = std::decay_t<E>>
 using names_t = decltype((names_v<D>));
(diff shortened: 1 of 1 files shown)


Neargye:
> detail::min_v<E> with E
It's a good catch, there really should have been a D, it was a typo.

In general, I like the changes, I will look at them a little more before the merge.

include/magic_enum.hpp
```diff
@@ -545,7 +547,7 @@ constexpr auto entries(std::index_sequence<I...>) noexcept {
 }
 
 template <typename E>
-inline constexpr auto entries_v = entries<E>(std::make_index_sequence<count_v<E>>{});
+inline constexpr std::array entries_v = entries<E>(std::make_index_sequence<count_v<E>>{});
```
Neargye, line 552:
```cpp
template <typename E>
inline constexpr auto names_v = names<E>(std::make_index_sequence<count_v<E>>{});
```
not need to change to `std::arrray`?

schaumb: I tried to compile only the examples and the tests. It runs without the `names_v` resolution. Probably not used directly anywhere.

But of course, it can be this too.

Neargye:
Corrected the style and made the same check for n<E, V>(). Can you check if these changes are ok for you? and I think this PR is ready

schaumb:
These changes are ok! Thanks :)

Neargye (approved):

Neargye:
Thanks!

JuliaGut:
@schaumb I really like the Magic Enum library, but unfortunately I only have a gcc8 compiler available for my project.
Do you know, if it is possible to use the function "magic_enum::enum_name" with gcc8 compiler? If so, could you tell me which version of the library to use for it?

schaumb:
Hi @JuliaGut
So the Magic Enum lib is not supported on gcc-8, because the hack used in this lib is not available on gcc-8.

If you want to use it, you need to specify all member's names individually.
It can be done like [this](https://godbolt.org/z/odc6dcfoM)

```cpp
enum class EN {
    asd,
    fgh
};

template<>
constexpr customize_t enum_name<EN>(EN e) {
    switch(e) {
        case EN::asd: return "asd";
        case EN::fgh: return "fgh";
    }
    return invalid_tag;
}
}
```

JuliaGut:
Thanks a lot for the quick response! 🥇

PR #138, opened 2022-02-18
Simplier change only one enum instance name

schaumb:
If someone needs to change only one enum instance name, it is much easier with this modification.

With this, it is not necessary to write a function but enough a constant variable.

So
```
template <>
constexpr std::string_view magic_enum::customize::enum_name<Color>(Color value) noexcept {
  switch (value) {
    case Color::RED:
      return "red";
    default:
      return {};
  }
}
```

can be simplified to
```
template<>
constexpr std::string_view magic_enum::customize::enum_name_v<Color, Color::RED> = "red";
```

--- 

reorder `enum_name_v`, move to `customize::` + add test

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -148,7 +148,7 @@ constexpr string_view enum_name(E) noexcept {
 }
 
 // If need custom name only for one enum, add specialization enum_name_v for that instance.
-template <typename E, E V>
+template <auto V, typename E = decltype(V)>
 inline constexpr auto enum_name_v = detail::n<E, V>();
 
 } // namespace magic_enum::customize
@@ -380,7 +380,7 @@ template <typename E, auto V>
 constexpr bool is_valid() noexcept {
   static_assert(is_enum_v<E>, "magic_enum::detail::is_valid requires enum type.");
 
-  return customize::enum_name_v<E, static_cast<E>(V)>.size() != 0;
+  return customize::enum_name_v<static_cast<E>(V)>.size() != 0;
 }
 
 template <typename E, int O, bool IsFlags, typename U = std::underlying_type_t<E>>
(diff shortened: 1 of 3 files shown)


schaumb:
It seems [Visual Studio 2017](https://github.com/Neargye/magic_enum/runs/5248790375?check_suite_focus=true) does not support this "feature".

Neargye:
I will check vs2017 locally.

Neargye:
Wouldn't just a shorter entry work
```cpp
template <>
constexpr std::string_view magic_enum::customize::enum_name<Color>(Color value) noexcept {
  return value == Color::RED ? "red" : {};
}
```
I'm a bit concerned about creating two different customization points that might conflict, for example.

schaumb:
This customization point enables to "cancel" an enum value.
So:
```
enum class A {
    _1,
    _1_5, 
    _2, 
    _3
};

template<>
constexpr std::string_view magic_enum::customize::enum_name_v<A, A::_1_5> = "";

static_assert(magic_enum::enum_count<A>() == 3);
```

I think it can be change to not needed to add type:
```
template<>
constexpr std::string_view magic_enum::customize::enum_name_v<A::_1_5> = "";
```

Neargye:
to enables/disables, I would add another function like 

```cpp
template <>
constexpr std::string_view magic_enum::customize::enum_valid<Color>(Color value) noexcept {
  return value == Color::RED ? false : true;
}
```

perhaps it will be more readable.

Neargye:
I don’t have time to watch pr yet because of the need for relocation from Kyiv.

schaumb:
VS 2017 cannot enable to write namespace constexpr variable specialization, so:
```
template<>
constexpr std::string_view magic_enum::customize::enum_name_v<Color, Color::RED> = "red";
``` 
I need to replace to (only because VS2017)
```
template<>
constexpr std::string_view enum_name_v<Color, Color::RED> = "red";
}
```

I wrote another test for the "canceling". ~~This fails because~~

~~`enum_name_v<Binary, Binary::ONE>` overrides somehow `enum_name_v<MaxUsedAsInvalid, MaxUsedAsInvalid::ONE>` default value (again: only for VS2017)~~

I manage it, no other VS2017 issue

schaumb:
So I added a new test for enum value injection.
This technique enables the following:

```
enum class Numbers : int { one = 1, two, three };
constexpr auto NumberFour = static_cast<Numbers>(4);

template<>
constexpr std::string_view magic_enum::customize::enum_name_v<NumberFour> = "four";

static_assert(enum_count<Numbers>() == 4);
```

schaumb:
https://github.com/Neargye/magic_enum/pull/166 will be awsome instead of this

PR #140, opened 2022-02-18
Constexpr switch inside

schaumb:
currently for inside usage only

PR #142, opened 2022-02-19
Constexpr switch names

schaumb:
Based on https://github.com/Neargye/magic_enum/pull/140 and master.

`enum_cast` with `string_view` parameter and default predicate use constexpr switch on predefined string hashes. This makes reaching enum value constant time by name


schaumb:
~~current string hash solution uses crc.~~

~~This can be hacked:~~

```
enum class crc_hack {
    b5a7b602ab754d7ab30fb42c4fb28d82,
    d19f2e9e82d14b96be4fa12b8a27ee9f
};

static_assert(magic_enum::enum_cast<crc_hack>("d19f2e9e82d14b96be4fa12b8a27ee9f") ); // error: duplicate case value
```

I made a test for this crc hack. I created a secondary string hash for safety and a check after hash equality. This has to safe enough :)

schaumb:
@Neargye can you review it?

Neargye:
In general, it looks good, I may later correct the style somewhere, but this is not critical.

Neargye:
@schaumb  https://github.com/Neargye/magic_enum/pull/62 Remembering this problem, it didn't occur when using a hash?

schaumb:
> @schaumb #62 Remembering this problem, it didn't occur when using a hash?

It's working https://godbolt.org/z/9s41noGKn

Neargye:
Good!

Issue #143, opened 2022-02-19
make typesafe enum_fuse


schaumb:
Currently it can be write the following:

```
enum class C1 { a, b, c };
enum class C2 { a, b, c };
switch(enum_fuse(C1::a, C2::b)) {
case enum_fuse(C2::a, C1::b):
   // oops it handled
   break;
}
```

So enum fuse is not recognize the parameter types.

It can be made typesafe with the following trick:

```
template<typename ... enums>
constexpr auto enum_fuse(enums ... vars) {
    enum class Enum : std::uintmax_t {}
    std::uintmax_t val = old_enum_fuse(vars...);
    return static_cast<Enum>(val);
}
```

pavelkryukov:
Looks great – it constructs a Cartesian product enum of N enums in compile time! It should solve problems of fused enums # mismatch too. And we can remove `+ 1` trick.

pavelkryukov:
Sadly it looks like UB:

```
C++ standard 5.2.9 Static cast [expr.static.cast] paragraph 7

A value of integral or enumeration type can be explicitly converted to an enumeration type. The value is unchanged if the original value is within the range of the enumeration values (7.2). Otherwise, the resulting enumeration value is unspecified / undefined (since C++17).
```

Neargye:
This is a good idea, but then the switch on such an enum will be hit warning C4063.

pavelkryukov:
Was about to post the same:

```c++
/Users/runner/work/magic_enum/magic_enum/test/test.cpp:1033:10: error: case value not in enumerated type 'std::__1::optional<Enum>::value_type' (aka 'Enum') [-Werror,-Wswitch]
    case magic_enum::enum_fuse(Color::BLUE, Directions::Down).value():
         ^
```

@Neargye but it is not an UB up to your knowledge?

Neargye:
AIK, not UB, I investigate how suppress this warning.

pavelkryukov:
```diff
-    enum class Enum : std::uintmax_t {}
+    enum class Enum : std::uintmax_t;
```

This hides the fact enum has no values from compilers.

But Visual Studio does not like `switch/case` over incomplete enums. It is a different warning (C4064), maybe safer to be ignored/disabled than C4063.

PR #144, opened 2022-02-19
Fix docs

pavelkryukov:


Neargye:
After that #143, I thought, maybe we should not use enum_indexes? And to allow any enums value, just cast it to int.

Neargye:
Like
```cpp
namespace detail::fusion {

constexpr std::uintmax_t cantor_pair(std::uintmax_t v1, std::uintmax_t v2) noexcept {
  return (((v1 + v2) * (v1 + v2 + 1)) >> 1) + v2;
}

template <typename... Ts>
constexpr std::uintmax_t cantor_pair(size_t v1, size_t head, Ts... tail) noexcept {
  return cantor_pair(cantor_pair(v1, head), tail...);
}

} // namespace magic_enum::detail::fusion

// Returns a bijective mix of several enum values. This can be used to emulate 2D switch/case statements.
template <typename... Es>
[[nodiscard]] constexpr auto enum_fuse(Es... values) noexcept {
  static_assert((std::is_enum_v<std::decay_t<Es>> && ...), "magic_enum::enum_fuse requires enum type.");
  static_assert(sizeof...(Es) >= 2, "magic_enum::enum_fuse requires at least 2 enums");
  static_assert((detail::log2(enum_count<Es>() + 1) + ...) <= (sizeof(std::uintmax_t) * 8), "magic_enum::enum_fuse does not work for large enums");
  enum class enum_fuse_t : std::uintmax_t {};
  const auto fuse = detail::fusion::cantor_pair(static_cast<std::uintmax_t>(values)...);

  return static_cast<enum_fuse_t>(fuse);
}
```

pavelkryukov:
1. What about negative values? Neither Cantor pairing nor OR pairing handles them correctly.
2. Cantor pairing was good for PoC, but it is slow and error-prone (multiplication overflow)

Neargye:
1. We cast to unit, so bo problem. Uint overflow not UB. 
2. I use cantor pair for example, we can use in final hash, etc.

pavelkryukov:
> We cast to unit, so bo problem. Uint overflow not UB.

Mutliplication overflow is wrong result even w/o UB, so let's forget about it.

For OR, you need to trim the negative value first:

```c++
enum Enum {
    MINUS_ONE = -1,
    MINUS_TWO = -2,
    ONE = 1;
    TWO = 2;
};

auto hash(E value, std::unitmax_t hash)
{
    size_t size = log2(enum_count<E>() + 1);
    hash <<= size;
    hash |= enum_value(value) & ((1ULL << size) - 1);
}
```

Finally, "0" is a special value to handle type/count mismatches. If we have type safety, it can be removed, but we have not achieved this point yet.

pavelkryukov:
Does MagicEnum assert `enum_value(x) < enum_count<E>() && enum_value(x) > -enum_count<E>()` ?

Neargye:
Enum_value return value from index, so assert if index > size.

Maybe you about enum_integer? It's  only cast to integer, without asset.

pavelkryukov:
Yes, confused these two. A clarified question would be:

In the code you provided, does `static_cast` return a result within `enum_count<E>`, so it can be safely mapped to log2(enum_count<E> + 1) bits?
```c++
  const auto fuse = detail::fusion::cantor_pair(static_cast<std::uintmax_t>(values)...);
```

Neargye:
It's only safe if enum in range.

pavelkryukov:
So, there may be unsupported input and we must keep `std::optional`.

Neargye:
> So, there may be unsupported input and we must keep `std::optional`. 

Look so, yes.

pavelkryukov:
Then we need `std::optional` in docs.

Neargye:
Need change to template <typename... Es>
[[nodiscard]] constexpr optional\<std::uintmax_t\> enum_fuse(Es... values);

pavelkryukov:
done.

Neargye:
Thanks

PR #145, opened 2022-02-19
Make enum_fuse typesafe (fixes #143)

### LINKED ISSUE #143: make typesafe enum_fuse

pavelkryukov:


pavelkryukov:
Added a chicken bit to disable type safety

doc/reference.md
```diff
@@ -330,17 +330,24 @@ constexpr string_view enum_type_name() noexcept;
 
 ```cpp
 template <typename... Es>
-[[nodiscard]] constexpr optional<std::uintmax_t> enum_fuse(Es... values);
+[[nodiscard]] constexpr std::optional<FusedEnum> enum_fuse(Es... values);
```
schaumb: optional not necessary means std::optional.

schaumb (changes requested):

doc/reference.md
pavelkryukov: fixed.

PR #146, opened 2022-02-19
Convert std::enable_if_t to static_assert for cleaner error message

pavelkryukov:
Prepare for less trivial type deduction


Neargye:
Thanks

ZXShady:
This breaks API though SFINAE was noticiable and you hid it now.

```cpp
requires {
   enum_fuse(e); // is no longer correct
};
```

Issue #147, opened 2022-02-27
fusion_detail doesn't handle MAGIC_ENUM_USING_ALIAS_OPTIONAL


l16i:
The current implementation of the `fusion_detail` namespace is

``` c++
namespace fusion_detail {

template <typename E>
constexpr std::optional<std::uintmax_t> fuse_one_enum(std::optional<std::uintmax_t> hash, E value) noexcept {
  if (hash.has_value()) {
    if (const auto index = enum_index(value); index.has_value()) {
      // Add 1 to prevent matching 2D fusions with 3D fusions etc.
      return (hash.value() << detail::log2(enum_count<E>() + 1)) | (index.value() + 1);
    }
  }
  return std::nullopt;
}

template <typename E>
constexpr std::optional<std::uintmax_t> fuse_enum(E value) noexcept {
  return fuse_one_enum(0, value);
}

template <typename E, typename... Es>
constexpr std::optional<std::uintmax_t> fuse_enum(E head, Es... tail) noexcept {
  return fuse_one_enum(fuse_enum(tail...), head);
}

}
```

It still uses `std::optional` instead of `optional`

Neargye:
please check master version

PR #154, opened 2022-03-04
remove std:: from optionals to enable again MAGIC_ENUM_USING_ALIAS_OPTIONAL 

schaumb:
https://github.com/Neargye/magic_enum/issues/147

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -914,7 +914,7 @@ template <typename E, typename BinaryPredicate = std::equal_to<char>>
 namespace fusion_detail {
 
 template <typename E>
-constexpr std::optional<std::uintmax_t> fuse_one_enum(std::optional<std::uintmax_t> hash, E value) noexcept {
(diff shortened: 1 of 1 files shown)


Neargye:
`return std::nullopt;` -> `return {};`

Neargye (approved):

Neargye:
thanks

schaumb:
@Neargye Does it need a test?

```
template<typename T>
struct MyOpt {
    constexpr MyOpt() : mHasValue{} {}
    constexpr MyOpt(const T& v) : mHasValue{true}, val{v} { }
    constexpr bool has_value() const noexcept { return mHasValue; }
    constexpr const T& value() const noexcept { return val; }
    bool mHasValue;
    T val;
};

#define MAGIC_ENUM_USING_ALIAS_OPTIONAL template<typename T> using optional = ::MyOpt<T>;

...
```

Neargye:
@schaumb I would create a separate test for this(and issues for tracking)

PR #159, opened 2022-03-08
add compile time enum_index

### LINKED ISSUE #158: add enum_index_v compile time checker for template constant enum value
I want to get enum value index.

Currently there is a `magic_enum::enum_index(enum_val)` which returns `optional`.
If I want to make a `magic_enum::enum_index(enum_val).value()`, on constant time there are no check for enum existing, runtime exception throws.

I need a `magic_enum::enum_index_v<enum_val>` constant which check `enum_val` existance compile time.

So this throw 'std::bad_optional_access' exception.
```
#include "magic_enum.hpp"
#include <iostream>

enum class A {
a1, a2, a3
};

template<A val>
void print_index() {
    std::cout << magic_enum::enum_index(val).value() << std::endl;
}

int main() {
    constexpr A known_value = static_cast<A>(5);
    
    print_index<known_value>();
}
```

If I wrote the following, compile time error happens:
```
#include "magic_enum.hpp"
#include <iostream>

namespace magic_enum {
    template<auto V, typename = decltype(V)>
    constexpr auto enum_index_v = magic_enum::enum_index(V).value();
}

enum class A {
a1, a2, a3
};

template<A val>
void print_index() {
    std::cout << magic_enum::enum_index_v<val> << std::endl;
}

int main() {
    constexpr A known_value = static_cast<A>(5);
    
    print_index<known_value>();
}
```

Neargye:


Neargye:
@schaumb could you please take a look?

doc/reference.md
```diff
@@ -275,8 +291,24 @@ constexpr optional<size_t> enum_index() noexcept;
...
   constexpr auto color_index = magic_enum::enum_index(Color::BLUE);
-  // color_index -> color_index.value() -> 1
-  // color_index -> color_index.has_value() -> true
+  // color_index.value() -> 1
+  // color_index.has_value() -> true
+  ```
+
+```cpp
+template <auto V>
+constexpr size_t enum_index() noexcept;
```
schaumb: Just a stylistic inconsistency:
At `enum_name` the documentation of the function variants was merged. These variants (`enum_index` and `enum_value`) can be merged too.

schaumb: documentation bug: [Missing the parameter](https://github.com/Neargye/magic_enum/blob/04a3d32d0fca3d3c79ae4ffde0f002bb4eb38bf8/doc/reference.md?plain=1#L267) from the other variant of `enum_index`

include/magic_enum.hpp
```diff
@@ -705,7 +708,7 @@ template <typename E>
 template <auto V>
 [[nodiscard]] constexpr auto enum_name() noexcept -> detail::enable_if_enum_t<decltype(V), string_view> {
   constexpr string_view name = detail::enum_name_v<std::decay_t<decltype(V)>, V>;
-  static_assert(name.size() > 0, "Enum value does not have a name.");
+  static_assert(name.size() > 0, "magic_enum::enum_name enum value does not have a name.");
```
schaumb: [clang-tidy says](https://clang.llvm.org/extra/clang-tidy/checks/readability-container-size-empty.html) use `!name.empty()` instead of `name.size() > 0`

Neargye: fixed, thanks

schaumb (approved):
Looks good, thanks!

PR #160, opened 2022-03-09
Fix optional masos

### LINKED ISSUE #155: Error message in macOS with latest version
When I compile with macOS XCode, the following message occurs on line 921:

```c++
//error: 'value' is unavailable: introduced in macOS 10.13
return (hash.value() << detail::log2(enum_count<E>() + 1)) | (index.value() + 1);
```

I can increase my minimum target to macOS 10.13 and presumably make it work, but I would like not to have to do that. Is there some way this code could be made optional? For now I am sticking with an earlier branch of `magic_enum`.

Neargye:


Neargye:
@rpatters1 could you please check this thread? (unfortunately, my mac remained in Kyiv)

rpatters1:
Yes, this resolves my issue. I hope you are staying safe.

Neargye:
Thanks for help

PR #161, opened 2022-03-10
Add support for the ROS build system

Timple:
This allows all ROS packages to include this wonderful library by listing this github repository as a dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,8 @@ write_basic_package_version_file(${PROJECT_NAME}ConfigVersion.cmake
         ARCH_INDEPENDENT)
 
 if(MAGIC_ENUM_OPT_INSTALL)
+    include(GNUInstallDirs)
+
     install(TARGETS ${PROJECT_NAME}
             EXPORT ${PROJECT_NAME}Config)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,14 +58,16 @@ endif()
 find_package(roscpp QUIET)
 find_package(rclcpp QUIET)
 if(roscpp_FOUND OR rclcpp_FOUND)
+        include(GNUInstallDirs)
+
         install(DIRECTORY include/
-                DESTINATION include/${PROJECT_NAME})
+                DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
         install(TARGETS ${PROJECT_NAME}
-                DESTINATION lib/${PROJECT_NAME})
+                LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
-        install(FILES package.xml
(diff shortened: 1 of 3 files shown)


Neargye:
Add notice to readme too.

Timple:
Marking as draft as I found a possible improvement.

Timple:
Works for ROS1.
ROS2 pending...

Timple:
ROS2 operational. Required an extra `target_link_libraries` step.

Neargye:
@Timple is it ready?

Timple:
Yes, verified on both ROS versions

Sorry for the iterations, I thought I was there on the first commit 🙂

Neargye:
Many iterations are ok.
thanks for pr!

Timple:
Thank you!

Would you mind if I release it to the ROS package manager as well?
Source and documentation will keep pointing to this repository.

Neargye:
I am for any improvement for the community.

As I understand, I need to update every release package.xml?

Timple:
Yes, keeping the versions in sync would be ideal! Similar to meson.build I guess...

Unfortunately the XML structure cannot parse git tags...

Neargye:
@Timple is it possible remove me from mailing in build system ros?

Timple:
Ow, I'm not sure how you auto-subscribed (and I did not). 

Seems to be because of the `<maintainer` taq in the `package.xml` file.

Shall I list myself as ROS maintainer as I'm updating the ROS buildfarm? 
I would list you as author, to give credit where it's due. Being an author does not auto-subscribe to the buildfarm mailings.

tdenewiler:
I was able to quickly add this to a ROS 2-based node I am developing. It's great! Thanks for the library and for the ROS integration.

One thing that would make this better is if I could resolve names of constants in ROS messages. In ROS a message is defined in a file like this:

```
int8 NAME_A = 0
int8 NAME_B = 1

int8 data
```

The ROS build system creates auto-generated C++ header files. The relevant contents look like this:

```
// constant declarations
  static constexpr int8_t NAME_A =
    0;
  static constexpr int8_t NAME_B =
    1;
// constant definitions
#if __cplusplus < 201703L
// static constexpr member variable definitions are only needed in C++14 and below, deprecated in C++17
template<typename ContainerAllocator>
constexpr int8_t NAME_A_<ContainerAllocator>::CONTINGENCY;
#endif  // __cplusplus < 201703L
#if __cplusplus < 201703L
// static constexpr member variable definitions are only needed in C++14 and below, deprecated in C++17
template<typename ContainerAllocator>
constexpr int8_t NAME_B_<ContainerAllocator>::CONTINGENCY;
#endif  // __cplusplus < 201703L
```

Do either of you know if I can use this library to resolve the names of the constants from values in this context?

PR #166, opened 2022-03-11
new customize

### LINKED ISSUE #153: create enable/disable customization point for enum values
@Neargye idea for the [enum disable feature](https://github.com/Neargye/magic_enum/pull/138#issuecomment-1057295127) is to create another customization point:

```
template <>
constexpr bool magic_enum::customize::enum_valid<Color>(Color value) noexcept {
  return value == Color::RED ? false : true;
}
```

This needs an implementation

Neargye:
new customize to fix https://github.com/Neargye/magic_enum/issues/153 and replace https://github.com/Neargye/magic_enum/pull/138

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -373,8 +373,10 @@ constexpr auto n() noexcept {
   using std::get;
 
   [[maybe_unused]] constexpr auto custom = customize::enum_type_name<E>();
+  static_assert(std::is_same_v<std::decay_t<decltype(custom)>, customize::customize_t>, "magic_enum::customize requires customize_t type.");
   if constexpr (holds_alternative<string_view>(custom)) {
     constexpr auto name = get<string_view>(custom);
+    static_assert(!name.empty(), "magic_enum::customize requires not empty string.");
     return static_string<name.size()>{name};
   } else if constexpr (holds_alternative<customize_default_tag>(custom) && supported<E>::value) {
 #if defined(__clang__) || defined(__GNUC__)
@@ -400,8 +402,10 @@ constexpr auto n() noexcept {
   using std::get;
 
   [[maybe_unused]] constexpr auto custom = customize::enum_name<E>(V);
+  static_assert(std::is_same_v<std::decay_t<decltype(custom)>, customize::customize_t>, "magic_enum::customize requires customize_t type.");
   if constexpr (holds_alternative<string_view>(custom)) {
     constexpr auto name = get<string_view>(custom);
+    static_assert(!name.empty(), "magic_enum::customize requires not empty string.");
     return static_string<name.size()>{name};
   } else if constexpr (holds_alternative<customize_default_tag>(custom) && supported<E>::value) {
 #if defined(__clang__) || defined(__GNUC__)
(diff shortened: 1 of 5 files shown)


Neargye:
draft of the idea, still thinking about how to make it better and about the names
@schaumb Please take a look when you have time.

include/magic_enum.hpp
```diff
@@ -342,8 +369,14 @@ inline constexpr bool is_enum_v = std::is_enum_v<T> && std::is_same_v<T, std::de
 template <typename E>
 constexpr auto n() noexcept {
   static_assert(is_enum_v<E>, "magic_enum::detail::n requires enum type.");
+  using std::holds_alternative;
```
schaumb: this `holds_alternative` will work with `MAGIC_ENUM_USING_ALIAS_VARIANT` ?
I think it safer to check the [`variant::index`](https://en.cppreference.com/w/cpp/utility/variant/index) constexpr function result.

include/magic_enum.hpp
```diff
@@ -363,23 +396,24 @@ inline constexpr auto type_name_v = n<E>();
 template <typename E, E V>
 constexpr auto n() noexcept {
   static_assert(is_enum_v<E>, "magic_enum::detail::n requires enum type.");
-  [[maybe_unused]] constexpr auto custom_name = customize::enum_name<E>(V);
+  using std::holds_alternative;
+  using std::get;
 
-  if constexpr (custom_name.empty()) {
-    if constexpr (supported<E>::value) {
+  [[maybe_unused]] constexpr auto custom = customize::enum_name<E>(V);
```
schaumb, line 394: probably a static_assert check is necessary here to make sure to exactly return with the expected variant type.
`static_assert(std::is_same_v<std::decay_t<decltype(custom)>, customize::customize_t>)`

or backward compatible way, check `std::string_view` too.

include/magic_enum.hpp
```diff
@@ -363,23 +396,24 @@ inline constexpr auto type_name_v = n<E>();
...
+  if constexpr (holds_alternative<string_view>(custom)) {
+    constexpr auto name = get<string_view>(custom);
+    return static_string<name.size()>{name};
```
schaumb, line 399: I think before this line need a `static_assert` too if the function returns with empty string.
The message is something like: `"Empty string result is not allowed here. Please return with default_tag or invalid_tag"`

test/test.cpp
```diff
@@ -35,6 +35,15 @@
 #include <sstream>
 
 enum class Color { RED = -12, GREEN = 7, BLUE = 15 };
+template <>
+constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<Color>(Color value) noexcept {
+  switch (value) {
+    case Color::RED:
+      return "red";
+    default:
+      return magic_enum::customize::default_tag;
```
schaumb: I think the namespace is not necessary if the function is in the same namespace.

schaumb:
Do we need check, that 2 enum value can not have the same name?

The current enum name switch can not handle this case.

include/magic_enum.hpp
Neargye: make sense,yes

test/test.cpp
schaumb: this function return expression in test.cpp was not changed to default_tag :)

Neargye:
remove alias variant
1. variant using only in detail
2. variant have to much interfaces(stl/boost/etc), boost not have holds_alternative  nor [variant::index](https://en.cppreference.com/w/cpp/utility/variant/index)

schaumb:
> Do we need check, that 2 enum value can not have the same name?
> 
> The current enum name switch can not handle this case.

I manage myself this question to https://github.com/Neargye/magic_enum/pull/168

Neargye:
> > Do we need check, that 2 enum value can not have the same name?
> > The current enum name switch can not handle this case.
> 
> I manage myself this question to #168

I will try to do this separately

PR #168, opened 2022-03-15
Enum switch

schaumb:


example/example_switch.cpp
```diff
@@ -0,0 +1,111 @@
...
+constexpr std::string_view DoWork() {
+    return "default";
+}
+
+template <>
+constexpr std::string_view DoWork<Color::GREEN>() {
+    return "override";
+}
+
+struct SwitcherLambda1 {
```
Neargye: Maybe use 
```
// helper type for the visitor #4
template<class... Ts> struct overloaded : Ts... { using Ts::operator()...; };
// explicit deduction guide (not needed as of C++20)
template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
```
For an example, this would be cleaner.

schaumb: I replace it, thanks!

Neargye:
On gcc7.5 not work https://godbolt.org/z/vf9K5xWds

schaumb:
> On gcc7.5 not work https://godbolt.org/z/vf9K5xWds

with customizing enum names, it works
```
    template<>
    constexpr std::string_view enum_name<Color>(Color c) noexcept {
        switch (c) {
            case Color::RED: return "RED";
            case Color::BLUE: return "BLUE";
            case Color::GREEN: return "GREEN";
        }
        return "";
    }
}
```

But this does not work because the code assumpts that enum has at least one element. 
I try to fix that. Probably need a test for this, with the latest compilers failing too on empty enums.

schaumb:
@Neargye  I created a separate [PR](https://github.com/Neargye/magic_enum/pull/169) for the gcc7.5 issue

Neargye:
It's ready to merge?

schaumb:
> It's ready to merge?

@Neargye  there are some missing documentation part, but the code is functionally done

schaumb:
Probably I am not the perfect person to write these documentation, my english is not that good.

PR #169, opened 2022-03-17
Sparse only if have an element. Switch only if -.

schaumb:


include/magic_enum.hpp
```diff
@@ -1064,7 +1064,7 @@ template <typename E, typename BinaryPredicate = std::equal_to<char>>
       return static_cast<D>(result);
     }
     return {}; // Invalid value or out of range.
-  } else {
+  } else if constexpr (detail::count_v<D> > 0) {
```
Neargye, line 1067: same for enum_index?

schaumb: Enum index not requires it because empty enum is not sparse.

schaumb: Probably I thought wrong that this will be OK.
So on contiguous enum the functions work on min_v/max_v.
At empty enum min_v is equal to max_v , but it is undefined (probably the first/last inspected value).
The functions misinterpret the parameters on empty enum, so that does some undefined behaviour.

Empty enum must be sparse. So:

I need to refactor constexpr cases array generator
I need to wrap all function which uses constexpr_switch

PR #176, opened 2022-03-31
fix: GCC warning [-Werror=sign-conversion]

### LINKED ISSUE #175: GCC: conversion to unsigned int warning ( [-Werror=sign-conversion])
```
third_party_deps/magic_enum-src/include/magic_enum.hpp:733:43: error: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
  733 |       crc = (crc >> 8) ^ crc_table[(crc ^ c) & 0xff];
      |                                           ^
```

`c` is signed, but `crc` is unsigned, the conversion is ok, but can you please make it explicit so we don't have to disable this useful GCC warning, locally or globally, just because of this?

```
      crc = (crc >> 8) ^ crc_table[(crc ^ static_cast<std::uint32_t>(c)) & 0xff];
```

abdes:
Resolves: #175


Neargye:
Thanks

Issue #151, opened 2022-03-04
std::formatter


tomjohnstone:
Hi,

Is it possible to define a std::formatter (for use with std::format) that will work on any enum value?

Thanks,
Tom

schaumb:
https://godbolt.org/z/3T8dq56Mz

It is possible :) 

What do you think @Neargye ?

Neargye:
It looks interesting.

tomjohnstone:
Unfortunately the enum value passed into the formatter needs to be a constexpr or else you get a compiler error, when using MSVC

Neargye:
Could you show example and the text of the error? @tomjohnstone

tomjohnstone:
Apologies, I went back to create a repro and it's working now. I must have made a copy and paste error when I tried putting your code into my codebase. Thank you very much @schaumb !

Neargye:
add in master, will be in v0.8.1 or v0.8.2

davidchisnall:
I have been unable to make this work with clang 16.0.0.6 and the associated version of libc++.  Is there any documentation on how it should be used (it's entirely possible that I'm holding it wrong)?

Neargye:
@davidchisnall 

Will update doc, for now short example:
```cpp
#include <magic_enum_format.hpp>

TEST_CASE("format-base") {
  REQUIRE(std::format("{}", Color::RED) == "red");
  REQUIRE(std::format("{}", Color{0}) == "0");
}
```

davidchisnall:
Thanks.  Turning this into a stand-alone program that I can test, I get:

```c++
#define _LIBCPP_ENABLE_EXPERIMENTAL
#include <magic_enum_format.hpp>
#include <format>
#include <iostream>

enum Color
{
	RED=1,
	GREEN,
	BLUE
};

int main()
{
	std::cout << std::format("{}", Color::RED) << std::endl;
}
```

Compiling this with clang 16.0.0.6  and the matched libc++ (the first `#define` is necessary to enable `std::format`), I get a lot of long compile errors:

```
In file included from enum.cc:3:
In file included from /usr/include/c++/v1/format:188:
/usr/include/c++/v1/__format/format_functions.h:354:7: error: call to deleted function '__determine_arg_t'
      __format::__determine_arg_t<_Context, remove_cvref_t<_Args>>()...};
[...]

PR #185, opened 2022-05-13
Magic enum format in a separated header file 

schaumb:
Based on https://github.com/Neargye/magic_enum/issues/151

### DIFF
--- a/include/magic_enum_format.hpp
+++ b/include/magic_enum_format.hpp
@@ -1,0 +1,74 @@
+//  __  __             _        ______                          _____
+// |  \/  |           (_)      |  ____|                        / ____|_     _
+// | \  / | __ _  __ _ _  ___  | |__   _ __  _   _ _ __ ___   | |   _| |_ _| |_
+// | |\/| |/ _` |/ _` | |/ __| |  __| | '_ \| | | | '_ ` _ \  | |  |_   _|_   _|
+// | |  | | (_| | (_| | | (__  | |____| | | | |_| | | | | | | | |____|_|   |_|
+// |_|  |_|\__,_|\__, |_|\___| |______|_| |_|\__,_|_| |_| |_|  \_____|
+//                __/ | https://github.com/Neargye/magic_enum
+//               |___/  version 0.8.0
+//
+// Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2019 - 2022 Daniil Goncharov <neargye@gmail.com>.
+//
+// Permission is hereby  granted, free of charge, to any  person obtaining a copy
+// of this software and associated  documentation files (the "Software"), to deal
+// in the Software  without restriction, including without  limitation the rights
+// to  use, copy,  modify, merge,  publish, distribute,  sublicense, and/or  sell
+// copies  of  the Software,  and  to  permit persons  to  whom  the Software  is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE  IS PROVIDED "AS  IS", WITHOUT WARRANTY  OF ANY KIND,  EXPRESS OR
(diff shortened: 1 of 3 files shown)


schaumb:
It needs some documentation (?) and test(s) ✓.

Neargye:
It looks really good with separate headers.

schaumb:
!!! This formatter uses std::string_view . If someone defines a different string_view, this will not work.

To resolve this problem one of the following is necessary:
- apply some conversion to std::string_view ✓ (`.data()`, `.size()` )
- ~~require its own formatter for the user-defined string_view~~ ✗
- ~~prohibit to use of this header~~ ✗

Neargye:
Is it possible to use a  char* formatter?

schaumb:
!!! Another problem is that this solution does not handle the enum flags.
- ~~one solution would be a unique format character, which enables to format of enum flags output, like `{...f}` ( 'f' character on the format string end can be a mean that handles this enum as a flag).~~ ✗
- or this can be the default behavior if the enum is a flag ✓

schaumb:
> Is it possible to use a char* formatter?

string_view data is not required to end with `'\0'`, so the `const char*` formatter cannot be used

include/magic_enum_format.hpp
```diff
@@ -0,0 +1,74 @@
...
+  constexpr bool enum_format_enabled() noexcept {
+    return MAGIC_ENUM_DEFAULT_ENABLE_ENUM_FORMAT;
+  }
+} // magic_enum::customize
+
+#include <format>
+
+template<typename E>
+struct std::formatter<E, std::enable_if_t<std::is_enum_v<E> && magic_enum::customize::enum_format_enabled<E>(), char>> : std::formatter<std::string_view, char> {
+  auto format(E e, format_context& ctx) {
```
Neargye, line 58: static_assert(std::is_constructible<std::string_view, decltype(magic_enum::enum_name(e)));
to fix https://github.com/Neargye/magic_enum/pull/185#issuecomment-1126173671

include/magic_enum_format.hpp
```diff
@@ -0,0 +1,74 @@
...
+    if (auto name = magic_enum::enum_name(e); name.empty()) {
+      constexpr auto type_name = magic_enum::enum_type_name<E>();
+      throw std::format_error("Type of " + std::string{type_name.data(), type_name.size()} + " enum value: " +
+            std::to_string(magic_enum::enum_integer(e)) + " is not exists.");
+    } else {
+      return this->std::formatter<std::string_view, char>::format(name, ctx);
```
Neargye: std::string_view{name.data(), name.size()}

include/magic_enum_format.hpp
schaumb: this check is not needed, because the other, `std::string_view{name.data(), name.size()}` conversion is enough

Timple:
Awesome job!

The [libfmt](https://fmt.dev/latest/api.html#format-api) library provides this functionality for `c++17`. This example works.

Would it make sense to combine this with this file?

```c++
template<typename E>
struct fmt::formatter<E, std::enable_if_t<std::is_enum_v<E>, char>> : fmt::formatter<std::string_view, char> {
  auto format(E e, format_context& ctx) {
    if constexpr (magic_enum::detail::is_flags_v<E>) {
      if (auto name = magic_enum::enum_flags_name(e); !name.empty()) {
        return this->fmt::formatter<std::string_view, char>::format(std::string_view{name.data(), name.size()}, ctx);
      }
    } else {
      if (auto name = magic_enum::enum_name(e); !name.empty()) {
        return this->fmt::formatter<std::string_view, char>::format(std::string_view{name.data(), name.size()}, ctx);
      }
    }
    constexpr auto type_name = magic_enum::enum_type_name<E>();
    throw std::domain_error("Type of " + std::string{type_name.data(), type_name.size()} + " enum value: " +
                            std::to_string(magic_enum::enum_integer(e)) + " is not exists.");
  }
};
```

Neargye:
@Timple please create new issues

PR #189, opened 2022-05-27
Fix compilation error, when including magic_enum after windows.h

bamo-gemini:
While the error is clearly windows's fault, windows.h is such a commonly used header, that I think it makes sense to support it. 
The fix is to wrap the function call to numeric_limits::max in braces, so the max-macro is avoided

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -739,7 +739,7 @@ struct constexpr_hash_t<Value, std::enable_if_t<std::is_same_v<Value, string_vie
     constexpr std::uint32_t operator()(string_view value) const noexcept {
       auto acc = static_cast<std::uint64_t>(2166136261ULL);
       for (const auto c : value) {
-        acc = (((acc ^ static_cast<std::uint64_t>(c)) * static_cast<std::uint64_t>(16777619ULL)) & std::numeric_limits<std::uint32_t>::max)();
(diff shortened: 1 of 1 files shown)


include/magic_enum.hpp
```diff
@@ -739,7 +739,7 @@ struct constexpr_hash_t<Value, std::enable_if_t<std::is_same_v<Value, string_vie
     constexpr std::uint32_t operator()(string_view value) const noexcept {
       auto acc = static_cast<std::uint64_t>(2166136261ULL);
       for (const auto c : value) {
-        acc = ((acc ^ static_cast<std::uint64_t>(c)) * static_cast<std::uint64_t>(16777619ULL)) & std::numeric_limits<std::uint32_t>::max();
+        acc = (((acc ^ static_cast<std::uint64_t>(c)) * static_cast<std::uint64_t>(16777619ULL)) & std::numeric_limits<std::uint32_t>::max)();
```
schaumb: I think here is not what it is meant to be.

`((acc ^ static_cast<std::uint64_t>(c)) * static_cast<std::uint64_t>(16777619ULL)) & (std::numeric_limits<std::uint32_t>::max)();`

the opening parenthesis need to be before `std::numeric_limits`

Neargye: `acc = ((acc ^ static_cast<std::uint64_t>(c)) * static_cast<std::uint64_t>(16777619ULL)) & (std::numeric_limits<std::uint32_t>::max)();`

Issue #190, opened 2022-05-27
signed/unsigned mismatch warning in visual studio 17.2.2 when enum_cast-ing integers to enum-values.


bamo-gemini:
The underlying type of the enum class is int.
See image below. Ask me for details, I can make a simple reproducable case if necessary.

![bilde](https://user-images.githubusercontent.com/35224692/170707469-1cb34b3b-cea5-4f78-962d-d7d106b9a2bc.png)

schaumb:
Hi @bamo-volue ,

Thanks for the issue report.
I can [reproduce](https://godbolt.org/z/czYbaMGn9) the problem. 
I will find a solution shortly.

PR #191, opened 2022-05-27
fix all VS /Wall warning

schaumb:
https://github.com/Neargye/magic_enum/issues/190

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -222,14 +222,14 @@ class static_string<0> {
 
   constexpr explicit static_string(string_view) noexcept {}
 
-  constexpr const char* data() const noexcept { return nullptr; }
+  [[maybe_unused]] [[nodiscard]] constexpr const char* data() const noexcept { return nullptr; }
 
-  constexpr std::size_t size() const noexcept { return 0; }
+  [[maybe_unused]] [[nodiscard]] constexpr std::size_t size() const noexcept { return 0; }
 
-  constexpr operator string_view() const noexcept { return {}; }
+  [[maybe_unused]] [[nodiscard]] constexpr operator string_view() const noexcept { return {}; }
 };
 
-constexpr string_view pretty_name(string_view name) noexcept {
+[[maybe_unused]] [[nodiscard]] constexpr string_view pretty_name(string_view name) noexcept {
   for (std::size_t i = name.size(); i > 0; --i) {
     if (!((name[i - 1] >= '0' && name[i - 1] <= '9') ||
           (name[i - 1] >= 'a' && name[i - 1] <= 'z') ||
@@ -256,7 +256,7 @@ constexpr string_view pretty_name(string_view name) noexcept {
 }
 
 class case_insensitive {
-  static constexpr char to_lower(char c) noexcept {
+  [[maybe_unused]] [[nodiscard]] static constexpr char to_lower(char c) noexcept {
     return (c >= 'A' && c <= 'Z') ? static_cast<char>(c + ('a' - 'A')) : c;
   }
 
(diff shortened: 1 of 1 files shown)


include/magic_enum.hpp
```diff
@@ -222,14 +222,14 @@ class static_string<0> {
 
   constexpr explicit static_string(string_view) noexcept {}
 
-  constexpr const char* data() const noexcept { return nullptr; }
+  [[maybe_unused]] [[nodiscard]] constexpr const char* data() const noexcept { return nullptr; }
```
Neargye, line 225: I'm a little surprised by this construction, as I know it contradicts each other.

schaumb: `[[nodiscard]]` means that if you use it, the return value cannot be discarded.
`[[maybe_unused]]` means that there is a chance to this function is not used.
These notations are not contradicting each other.

If someone did not use this function, vs will warn about it, this is why `[[maybe_unused]]` needs because no one is used usually. 
If any code calls this function, that will generate a warning if it discards the return value. So:
```
int main() {
  static_string<0> ss{""};
  ss.data(); // the return value not used, this will generate a warning.
}
```

Neargye: And what warnings does it fix?

Neargye: I mean, it's all internal function, do we real need this? @schaumb

schaumb: I use clion as the development environment and the built-in inspection tools recognize that these const member functions have no external effect, so it only makes sense to call because of the return value. If the return value is not used, it makes no sense to call the function. The `[[nodiscard]]` tag was automatically added because of this recommendation.

This tag can be omitted because it is not strictly part of the issue.

Neargye: I think [[maybe_unused]] may be omitted because it's internal stuff and I __hope__ we will use it right and without warnings.

schaumb: The `[[maybe_unused]]` tag fixes the following warnings:

```
https://raw.githubusercontent.com/Neargye/magic_enum/master/include/magic_enum.hpp(225): warning C4514: 'magic_enum::detail::static_string<0>::data': unreferenced inline function has been removed
https://raw.githubusercontent.com/Neargye/magic_enum/master/include/magic_enum.hpp(229): warning C4514: 'magic_enum::detail::static_string<0>::operator std::string_view': unreferenced inline function has been removed
https://raw.githubusercontent.com/Neargye/magic_enum/master/include/magic_enum.hpp(232): warning C4514: 'magic_enum::detail::pretty_name': unreferenced inline function has been removed
https://raw.githubusercontent.com/Neargye/magic_enum/master/include/magic_enum.hpp(259): warning C4514: 'magic_enum::detail::case_insensitive::to_lower': unreferenced inline function has been removed
https://raw.githubusercontent.com/Neargye/magic_enum/master/include/magic_enum.hpp(275): warning C4514: 'magic_enum::detail::find': unreferenced inline function has been removed

```

Neargye: WIll check

Neargye: I propose add `#pragma warning(disable : 4514) // Unreferenced inline function has been removed.` at line 73.
This may not be correct, but I prefer to silence warnings that don't seem right.

schaumb: The `[[maybe_unused]]` is standardized, and it's part of the C++ language. the `#pragma warning` is not. 
The warning has right, these functions are not used. I think this tag fits here, especially because these are internal stuff.

I can replace it in this PR, but I have feelings against it.

Neargye: But we use this functions)

schaumb: We don't use this function (based on the compiler).

These functions are `static_string<0>` members whose are only used if some enum has no name, or is not recognized.
`static_string<i>::data` is used only at `static_string<i>::operator string_view`, but `static_string<0>::operator string_view` returns a default constructed value, so nowhere used `data` and `size`.
 
The other 'free' functions are only used if the `magic_enum` specific functions are used, and template-specialized with some enum.

Issue #42, opened 2020-07-05
C++ language standard, issues with explicit setting to 17


davidhunter22:
I recently tried the library via vcpkg on MSVC. I had a problem that the following property is specified in magic_enumConfig.cmake

`INTERFACE_COMPILE_FEATURES "cxx_std_17"`

This is a problem as our code builds uses what MSVC calls "std::c++latest" which is newer than 17. Im not sure what should happen but it appears that the magic enum value overrides ours so we get downgraded to 17 which means a bunch of our code no longer compiles. 

In the nameof library you mention in issue https://github.com/Neargye/nameof/issues/12 that "I set c++17 for hand, because old cmake can't set c++17 for msvc++." I don't know if if makes sense for a package to require a specific version. If my projects use say c++ 11 and I try to include  magic_enum.hpp I get a very clear error message from the compiler telling me I need to use c++ 17.  I did look around vcpkg but couldn't find any documentation on best practice for settting c++ language standard.

I think what is really wanted is to require a minimum version but allow newer ones like latest or 20. I don't know how to do that in CMake/VCPKG.
For now we commented out the explicit setting of C++ 17 and things work great.

Neargye:
Thanks for your report. I will add this change in the next release.

PR #203, opened 2022-08-03
Add C++17 requirements in CMakeLists.txt

SijmenHuizenga:
Explicitly specify that this package requires c++17 language features.

After this change, building this package using a toolchain configured with a default cpp standard before 17, cmake understands that it should compile using cpp17.


Neargye:
https://github.com/Neargye/magic_enum/issues/42

PR #210, opened 2022-08-20
Typo fixed

nektonick:
MAGIC_ENUM_RANGE can't be less than UINT16_MIN since UINT16_MIN==0. So, this is typo.


Neargye:
Oh, thanks!

PR #212, opened 2022-09-02
New enum_switch

Neargye:


Neargye:
@schaumb Could you take a look, I don’t really understand what I could break with the hash?

PR #221, opened 2022-11-09
compile-time optimization

Neargye:


Neargye:
@Trass3r If it doesn't difficult, can you check this option? Do you have any other ideas of what can be optimized?

Neargye:
On clang compile time example
```cpp
#include <magic_enum.hpp>
int main(){}
```

Total Execution Time: 0.3125 -> 0.2656

Before first optimization 0.4062 -> 0.2656

PR #227, opened 2022-12-18
Speed up constexpr evaluation

AntonShalgachev:
The `pretty_name` function is called very frequently during the compilation, and with a modified clang I've noticed that this function can be slightly changed to speed up compilation. On my machine build time for `example.cpp` was reduced from 2.4s to 1.9s (-20%) with this change. Doesn't look like much, but this would save time for each unique enum used in the translation unit

Clang traces for the reference (can be viewed in [Perfetto](https://ui.perfetto.dev/)):
Before: [example.cpp-before.json.zip](https://github.com/Neargye/magic_enum/files/10254850/example.cpp-before.json.zip)
After: [example.cpp-after.json.zip](https://github.com/Neargye/magic_enum/files/10254848/example.cpp-after.json.zip)

Future ideas for further improvements:
* `std::string_view` seems to have a big overhead when being evaluated during the compilation (due to sanity checks at least on MSVC), so using a very simple custom string view class internally might be of help, preferably as a struct without member functions. In my project it reduced compilation time by another 30%
* I've experimented with a single function instantiation with a pack of N template value arguments (instead of N function instantiations with 1 template value argument each, where N is `max - min + 1`), and that resulted in a faster compilation in my case, so it's also something to consider (although this might be a very big change)

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -251,27 +251,34 @@ class static_string<0> {
 };
 
 constexpr string_view pretty_name(string_view name) noexcept {
+  char const* str = name.data();
   for (std::size_t i = name.size(); i > 0; --i) {
-    if (!((name[i - 1] >= '0' && name[i - 1] <= '9') ||
-          (name[i - 1] >= 'a' && name[i - 1] <= 'z') ||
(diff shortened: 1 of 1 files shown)


Neargye:
Thanks!

PR #237, opened 2023-01-31
Fix Clang 16 build in GitHub Actions

striezel:
The build for Clang 16 on Ubuntu is currently failing, see for example https://github.com/Neargye/magic_enum/actions/runs/4054633442/jobs/6976700211. The reason for that is a missing APT package repository for Clang 16 packages, so no `clang-16` package can be installed.

This pull requests adds the APT package repository for Clang 16 to fix that problem.


Neargye:
Thanks, when I added it, this package did not exist yet.

PR #238, opened 2023-02-01
Support ReSharper C++ builtins for type/enumerator name

Smertig:
In [ReSharper C++](https://www.jetbrains.com/resharper-cpp/) we have our own constexpr evaluation engine. However, getting enumerator or type name with `__PRETTY_FUNCTION__` hacks slows down perfomance of its on-the-fly code analysis. For example, resolving of `std::array<int, magic_enum::enum_count<SomeEnum>()>` requires bruteforcing of all enum names in range `[MAGIC_ENUM_RANGE_MIN, MAGIC_ENUM_RANGE_MAX]` leading to enormous number of template instantiations and constexpr calculations.

Since upcoming 2023.1 version ReSharper has [builtins](https://youtrack.jetbrains.com/issue/RSCPP-33856/Add-intrinsics-for-type-name-enumerator-name), that can be used to get type name or enumerator name directly: `__rscpp_type_name<T>()` and `__rscpp_enumerator_name(value)`. They greatly speed up evaluation of all `n<T>()` and `n<E, V>()` internal functions.

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -77,7 +77,7 @@
 #endif
 
 // Checks magic_enum compiler compatibility.
-#if defined(__clang__) && __clang_major__ >= 5 || defined(__GNUC__) && __GNUC__ >= 9 || defined(_MSC_VER) && _MSC_VER >= 1910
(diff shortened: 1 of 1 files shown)


Neargye:
Thanks!

PR #240, opened 2023-02-10
added -std=c++17 to flags so clangd would identify c++17 headers properly

embeddedpenguin:
I was having an issue where clangd would not properly see c++17 header files for magic_enum examples only. Clangd had no problem sorting through headers of files that were compiled with -std=c++17. Here's what I mean (from example/example.cpp):

image

This PR fixes that. This may be a change that doesn't need to happen, but I thought I would make the pr just in case.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,5 +1,6 @@
 include(CheckCXXCompilerFlag)
 
+set(CMAKE_CXX_STANDARD 17)
 if((CMAKE_CXX_COMPILER_ID MATCHES "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
     set(OPTIONS -Wall -Wextra -pedantic-errors -Werror)
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")


example/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 include(CheckCXXCompilerFlag)
 
 if((CMAKE_CXX_COMPILER_ID MATCHES "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
-    set(OPTIONS -Wall -Wextra -pedantic-errors -Werror)
+    set(OPTIONS -Wall -Wextra -pedantic-errors -Werror -std=c++17)
```
TheStormN: Perhaps a better approach would be to use `set(CMAKE_CXX_STANDARD 17)`.

embeddedpenguin: Probably, yeah. I cannot test it though due to #244

Neargye: @embeddedpenguin  https://github.com/Neargye/magic_enum/issues/244 fixed in master

embeddedpenguin: > Perhaps a better approach would be to use `set(CMAKE_CXX_STANDARD 17)`.

Updated to use `set(CMAKE_CXX_STANDARD 17)`.

Neargye (approved):

Neargye:
Thanks

PR #249, opened 2023-03-17
Fix std::forward variadic arguments in invoke_deoptimized()

aalexfvk:
Small fix. 
`std::forward<Args...>(args...)` is specialized to smth like `std::forward<Args_1, Args_2, ...>(args_1, args_2,...)`. But `std::forward` accepts only one argument.


Neargye:
Thanks for your improvements, but this applies to 3party libraries https://github.com/catchorg/Catch2
I would advise to make a PR in catch2, and I would later update the version of the dependencies

PR #251, opened 2023-03-24
allow special characters in enum, not only '_'

### LINKED ISSUE #241: magic_enum::enum_name doesn't return the right string if a prefix is a symbol
Hello,

I'm using a API rest and I've a enum class that match some of its fields. Some of them have a `$` prefix.

My enum look like this:
```c++
enum class MyEnum { 
$fieldOne,
$field_two
};
```

When I use the `magic_enum::enum_name` function the string is wrong and the prefix `$` is not here anymore. I get the following results:

```c++
const auto fieldOne = magic_enum::enum_name< MyEnum>(MyEnum::$fieldOne);
// "fieldOne" instead of "$fieldOne"

const auto fieldTwo = magic_enum::enum_name< MyEnum>(MyEnum::$fieldTwo);
// "field_two" instead of "$field_two"
```

Is it expected? Or did I do something wrong?

Thanks

IonutLascu:
closes https://github.com/Neargye/magic_enum/issues/241

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -263,7 +263,15 @@ class static_string<0> {
   constexpr operator string_view() const noexcept { return {}; }
 };
 
-constexpr string_view pretty_name(string_view name) noexcept {
+template <char... Cs>
+struct allowed_chars {
+  static constexpr bool is_allowed(char c) noexcept {
+    return ((c == Cs) || ...);
+  }
+};
+
+template <typename SpecialChars>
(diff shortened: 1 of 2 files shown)


IonutLascu:
Looks like `$` character is now allowed for all compilers, what you suggest to do @Neargye  ? I tried to extend your mechanism to allow more characters then `_`.

Neargye:
I would probably add a setting to add special characters

Neargye:
Fixed in master

PR #254, opened 2023-04-23
add bzlmod support

zaucy:
Adds support for [bazel modules](https://bazel.build/external/module) and prepares magic enum to be published to the [bazel central registry](https://github.com/bazelbuild/bazel-central-registry). When/if magic_enum gets published to the bazel central registry bazel users may depend on magic_enum simply by adding this to their `MODULE.bazel` file:

```py
bazel_dep(name = "magic_enum", version = "0.8.2")
```

I've also moved the targets for the examples into the example folder and moved the test targets into their own separate module. Tests being in their own separate module is recommended so that testing frameworks don't get published.

Additionally this PR adds support for magic_enum to compile on Windows with MSVC (i.e. using `/std:c++17` instead of `-std=c++17`) with bazel


.github/workflows/ubuntu.yml
```diff
@@ -52,13 +52,6 @@ jobs:
           sudo apt install ${{matrix.compiler.cxx}} -y
         fi
 
-    - name: Configure bazel
```
zaucy, line 55: This should be unnecessary for the runner images you are using. Bazelisk will utilise and install the bazel version specified by `.bazelversion` and is preinstalled!

example/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+    copts = COPTS,
+) for example in _EXAMPLES]
+
+cc_binary(
+    name = "example_nonascii_name",
+    srcs = ["example_nonascii_name.cpp"],
+    deps = ["@magic_enum"],
+    copts = COPTS,
+    target_compatible_with = select({
+        "@bazel_tools//tools/cpp:msvc": ["@platforms//:incompatible"],
```
zaucy: I'm not sure if the nonascii_name example is expected to work with MSVC, but for me it failed so I marked it as incompatible.

Neargye: Can we turn `nonascii_name` off by default?

zaucy: Certainly! I added a `"manual"` tag which means it may only be built explicitly i.e. `bazel run //example:example_nonascii_name`. Does that work for you?

Alternatively I could add a constraint which would require upstream users to declare they support  the nonascii feature in their platform definition. https://bazel.build/docs/configurable-attributes#platforms

Neargye (approved):

Neargye:
Thanks!

PR #255, opened 2023-04-24
meson: add options and tests

v1993:
Allows to enable flags from meson instead of relying on defines. Also support building/running tests from meson.


Neargye:
Could you add a short description of the meson to the readme? 
I myself do not use this build system and do not even know what to write.

v1993:
Not really sure what to write about it... Meson projects are most likely to use this via [meson's wrapdb](https://github.com/mesonbuild/wrapdb). Maybe simply add a badge like you do for vcpkg and the like? Something like

[Meson wrap](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/magic_enum.wrap)

Neargye:
yes, I think it will ok.

Neargye:
Thanks!

PR #258, opened 2023-04-29
Support gcc -fno-pretty-templates

ecatmur:
The gcc option [-fno-pretty-templates](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-fno-pretty-templates
) changes the `__PRETTY_FUNCTION__` from e.g. `"auto n() [with E = E]"` to `"auto n<E>()"` (a bit more like MSVC). Demo: https://godbolt.org/z/dj86T77b9

Construct a string_view from `__PRETTY_FUNCTION__` , and inspect the last character to determine how much to truncate before passing to `pretty_name`.

### DIFF
--- a/include/magic_enum.hpp
+++ b/include/magic_enum.hpp
@@ -266,6 +266,13 @@ class static_string<0> {
 
 constexpr string_view pretty_name(string_view name) noexcept {
   const char* str = name.data();
+#if defined(__clang__) // ]
+  name.remove_suffix(1);
+#elif defined(__GNUC__) // ] or >() (-fno-pretty-templates)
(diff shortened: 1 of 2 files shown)


include/magic_enum.hpp
```diff
@@ -266,6 +266,13 @@ class static_string<0> {
 
 constexpr string_view pretty_name(string_view name) noexcept {
   const char* str = name.data();
+#if defined(__clang__) // ]
```
Neargye: I understand that this will cause duplication code, but for now can you move this inside the n() function?

Neargye:
Thanks!

PR #262, opened 2023-05-12
Add gcc 12 and NONASCII flags to ubuntu build matrix.

talisein:
Without #261 the gcc-12 run will fail.


.github/workflows/windows.yml
```diff
@@ -37,9 +37,9 @@ jobs:
         mkdir build
         cd build
         cmake .. -A x64
-        cmake --build . --config Release
+        cmake --build . -j --config Release
```
Neargye: Isn't it necessary to write the number of threads after the parameter `j`?

talisein: The man page said it would default to the build tools default, but I forgot make would just become unbounded. I've added a parameter.

Neargye:
Thanks for all PR!

PR #277, opened 2023-06-02
add enum_flags_test and enum_flags_test_any functions for flags

Neargye:


Neargye:
@jonpetri Could you please review?
I change behavior for `test` func for 0, to be more consistent with Qflags or default test `if (flags & flag)`

PR #279, opened 2023-06-05
Re-enable Wenum-constexpr-conversion clang fix

rupprecht:
This was disabled in 427a47394f39ddb1517443acb808c1c861a35323, possibly by accident.


Neargye:
Thanks

PR #282, opened 2023-06-21
meson_options.txt: remove leftover noascii option

v1993:
Seems like this file was overlooked when purging noascii option from project.

I've also considered adding an option for recently introduced `MAGIC_ENUM_AUTO_IS_FLAGS`, but I'm not sure. If it's purely for compatibility reasons then it's probably not worth adding (and it's not exposed in CMake either). Would like to hear your opinion on this.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -6,13 +6,6 @@ option(
 )
 
 option(
-    'noascii',
-    type : 'boolean',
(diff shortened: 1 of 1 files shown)


Neargye:
Hi, thanks, I really forgot to delete.

Issue #288, opened 2023-08-04
Publish to bazel central registry


zaucy:
I started a draft to submit magic_enum to the BCR: https://github.com/bazelbuild/bazel-central-registry/pull/811

The current problem with it is they require an archive be released instead of relying on the generated one. Reasoning for that is here: https://blog.bazel.build/2023/02/15/github-archive-checksum.html. Could an archive be added to the release assets?

Also to publish automatically on release the [Publish to BCR](https://github.com/bazel-contrib/publish-to-bcr) GitHub app could be used.

Neargye:
Hi,
will take a look and try add archive

PR #294, opened 2023-09-11
automatically create and upload bzlmod archive

zaucy:
this will help with #288 by automatically making an archive available with a simple `git archive` command


Neargye:
Thanks!

PR #295, opened 2023-09-11
default bazel copts condition

zaucy:
from some feedback I got here: https://github.com/bazelbuild/bazel-central-registry/pull/811#discussion_r1302987328


Neargye:
Thanks!

PR #304, opened 2023-10-18
Remove superfluous semicolon

RaceTheMaSe:
Triggered when compiled with -Wpedantic -Werror and -Wextra-semi-stmt - results in:
empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]


Neargye:
Thanks!

PR #308, opened 2023-11-12
Fix magic_enum::containers::set erase function for Clang compilers

Ry-DS:
The erase function does not compile on the following:
```
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
```
When building for C++ 17 and 20 (note these specific configurations are all I tested).

Compile error:
```
include/magic_enum_containers.hpp:985:10: error: no viable conversion from 'typename container_type::reference' (aka 'reference_impl<>') to 'bool'
    bool res = ref;
         ^     ~~~
(retracted): note: in instantiation of member function 'magic_enum::containers::set<Retracted>::erase' requested here
        set.erase(Retracted::ELEMENT);
                         ^
include/magic_enum_containers.hpp:526:38: note: explicit conversion function is not a candidate
    [[nodiscard]] constexpr explicit operator bool() const noexcept { return (parent->a[num_index] & bit_index) > 0; }
```

This PR fixes that, while adding tests for this function to ensure this template is instansiated and exercised.

### DIFF
--- a/include/magic_enum/magic_enum_containers.hpp
+++ b/include/magic_enum/magic_enum_containers.hpp
@@ -982,7 +982,7 @@ class set {
 
   constexpr size_type erase(const key_type& key) noexcept {
     typename container_type::reference ref = a[key];
-    bool res = ref;
+    bool res = static_cast<bool>(ref);
     if (res) {
       --s;
     }
--- a/test/test_containers.cpp
+++ b/test/test_containers.cpp
@@ -270,6 +270,12 @@ TEST_CASE("containers_set") {
   REQUIRE_FALSE(color_set.empty());
   REQUIRE(color_set.size() == 3);
   REQUIRE(magic_enum::enum_count<Color>() == color_set.size());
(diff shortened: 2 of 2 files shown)


Neargye:
Thanks!

Ry-DS:
In hindsight, removing the explicit keyword here: https://github.com/Neargye/magic_enum/blob/fd9f70b91d04bf090699a4000b5d8b77781d125a/include/magic_enum_containers.hpp#L535
might have made more sense. Regardless either solution works, thanks for the prompt merge + release! :raised_hands:

Neargye:
Yes, perhaps this fix makes sense, in the original bitset it is not marked explicit  https://en.cppreference.com/w/cpp/utility/bitset/reference

If you have time, I will be glad to accept PR

Ry-DS:
was about to make one but I see it's handled in 677636ec430912e75f07554259de6331660d8fc7, thanks!

Issue #310, opened 2023-11-13
New include path format


cpsauer:
Hi @Neargye! Thanks again for a wonderful library.

I was looking at https://github.com/Neargye/magic_enum/commit/eb867ec9baaabc5b60816861923fdef53972cba2: is the intent to have users `#include "magic_enum/..."` (like for Bazel) or `#include "magic_enum.hpp"`, (like in the cmake config)?

Either seems totally reasonable, but I happened across the inconsistency when updating (and needing to change my includes), so I figured you might want a friendly heads.

Thanks for all you do,
Chris

Neargye:
Hi,
~~I think `#include "magic_enum.hpp"`, I'll try to quickly fix it.~~
~~Thanks for catch and create issues~~

UPD:
In https://github.com/Neargye/magic_enum/pull/374 was added breaking changes, need to use `#include <magic_enum/magic_enum.hpp>` instead of `#include <magic_enum.hpp>`

Neargye:
It seemed to me that I had configured this correctly in cmake this way `#include "magic_enum.hpp"`, but with so many build systems I’m already a little confused, could you tell me where I'm wrong?

cpsauer:
Sorry--I wrote the path <-> build systems mappings backwards above, but they are indeed inconsistent. (Have edited the original post so I don't confuse any future readers, but it was previously wrong.) 

So to confirm: cmake is as you want, `#include "magic_enum.hpp"`. I'm seeing you made the parallel change for meson in https://github.com/Neargye/magic_enum/commit/3c92e9035bbddd2a56eae09a04b2e9f6d59c1372.  I've tossed up a quick PR #311 to make things consistent for Bazel (which is what I'm using and why I noticed). On a quick look, it seemed like those were all the build systems; if that's right #311 should resolve!

Thanks again for making a stupendous library and sharing it with the world!

Neargye:
Thanks for your help!

cpsauer:
My pleasure. Thanks again for being so great to work with, Daniil, and for putting in the work to share such a great library with the world.

Neargye:
UPD:
In https://github.com/Neargye/magic_enum/pull/374 was added breaking changes, need to use `#include <magic_enum/magic_enum.hpp>` instead of `#include <magic_enum.hpp>`

PR #311, opened 2023-11-13
Bazel: Make include directories consistent

cpsauer:
Parallel to
https://github.com/Neargye/magic_enum/commit/3c92e9035bbddd2a56eae09a04b2e9f6d59c1372 https://github.com/Neargye/magic_enum/commit/eb867ec9baaabc5b60816861923fdef53972cba2

Addresses Bazel part of https://github.com/Neargye/magic_enum/issues/310

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -11,5 +11,5 @@ cc_library(
     name = "magic_enum",
     hdrs = glob(["include/magic_enum/*.hpp"]),
     copts = COPTS,
-    includes = ["include"],
+    includes = ["include/magic_enum"],
 )


zaucy:
Dang! Down side to the bazel CI be `continue-on-error`. Would've been caught before release. https://github.com/Neargye/magic_enum/actions/runs/6836609904/job/18591721634#step:5:46

zaucy (approved):

cpsauer:
^ Might it be worth changing that?

Also, if we can get to quick agreement on #312, we might want to skip merging this to avoid the unnecessary conflict. So maybe let's see if we can get #312 in by Wednesday and if not, we'll merge this?

Neargye:
Removed continue-on-error in CI

PR #312, opened 2023-11-14
Simplify Bazel

### LINKED ISSUE #309: Bazel: Might be worth removing dependency on rules_cc
Hey guys! First and foremost, thanks for a wonderful library.

I'm making use of it from bazel, and updating to 0.9.4, things broke with errors like `ERROR: no such package '@rules_cc//cc/compiler': BUILD file not found in directory 'cc/compiler' of external repository @rules_cc. Add a BUILD file to a directory to mark it as a package.`
(Copying that in so it's searchable by others.)

The issue is that some other (bazel first-party) repos (like skylib) bring in *super* old versions of rules_cc that don't have the compiler subdirectory, preempting this one's use of it. Most users, will, I think have some trouble tracking this down, since it's a bit tricky to figure out which transitive dependency calls pull it in. Plus, like rules_cc is explicitly not really intended for primetime use (per its readme). Those usages of it are, to the best of my knowledge, out-of-date holdovers from a time when bazel thought it was going to federate out more of its core.

I know that at some level that's not your problem. (And don't worry, I'll try to work to get those other repos fixed.) But I wonder if the most elegant solution might be to switch bazel/copts.bzl back to referencing the equivalent @bazel_tools//tools/cpp:(compiler) [values](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/BUILD.tools) rather than @rules_cc//cc/compiler, eliminating the unnecessary (and problematic) dependency on rules_cc. (Alternatively, we could add a transitive dependencies macro to bring in the rules_cc dependency for WORKSPACE users, but I think it's very likely to have been already shadowed by the time people use this.) What do you think? Tagging @zaucy, since I know he's been working on improvements here. Tagging also @fmeum, since I saw he'd given the feedback, and I've had lots of great interactions with him in the past :)

Thanks for your consideration,
Chris

cpsauer:
Please see #309 for context, but this sidesteps the bazel dependencies and cleans up the copts where they're not actually used.

Fixes #309


test/BUILD.bazel
```diff
@@ -1,23 +1,25 @@
...
 _TESTS = [
     "test",
     "test_flags",
 ]
 
+_MSVC_FLAGS = ["/std:c++17", "/permissive-"]
+_COPTS = select({
+    "//conditions:default": ["-std=c++17"],
+    "@bazel_tools//tools/cpp:clang-cl": _MSVC_FLAGS,
+    "@bazel_tools//tools/cpp:msvc": _MSVC_FLAGS,
```
zaucy: These options don't work unless specifying platforms as far as I know. Example from my previous bzlmod PR https://github.com/Neargye/magic_enum/pull/254/files#diff-e13b51fb7feeb38d3fb7dee88724107c57fb3d4805068b5f8d57a2d0900fb0dc. It'll just fallback to the default.

e.g.
```
cl : Command line warning D9002 : ignoring unknown option '-std=c++17'
```
Followed by many errors.

zaucy (changes requested):
If there's a way to not declare platforms in the test (and everyone upstream who uses bzlmod) that would be great.

test/BUILD.bazel
cpsauer: Ah shoot. Sorry I missed that. The other one worked without platforms, I take it? Presumably the toolchains set the @bazel_tools//tools/cpp:compiler flag internally?

I think the right move, then, is to re-add the bazelmod rules_cc dependency for tests (only). Seems cleaner than duplicating the config settings and still accomplishes all the previous goals around consuming the library itself. I'll push up a commit that does just that.

test/MODULE.bazel
```diff
@@ -1,6 +1,6 @@
 module(name = "magic_enum_tests")
 
 bazel_dep(name = "magic_enum")
-bazel_dep(name = "rules_cc", version = "0.0.8")
-
 local_path_override(module_name = "magic_enum", path = "..")
+
+bazel_dep(name = "rules_cc", version = "0.0.8")
```
zaucy, line 6: nit: maybe put the bazel dep back up top? :P

zaucy (approved):
Tested on one of my projects and it seems fine to me! :+1: (although I only tested on Windows)

test/MODULE.bazel
cpsauer: I'd reordered intentionally, thinking we might want the magic-enum dependency configuration all together for clarity that it's being overridden and quick-skimmability.

But I don't feel super strongly or anything--feel free to change if you feel strongly the other way? Or generally lmk if you think there's a different principle I should be absorbing here.

(I'm assuming from the approval that you don't want to block merging on this, yeah? Still v interested in what you have to say, ofc.)

cpsauer:
Thanks so much for testing @zaucy, and working so thoughtfully with me to make things even better. I really appreciate your being great to collaborate with!

All good to merge, I think, yeah?

@Neargye, thanks to you too for being so responsive and great!

Cheers,
Chris

Neargye:
@cpsauer  @zaucy Thanks for help!

cpsauer:
Thank *you*, Daniil!

cpsauer:
@Neargye, how imminent is v0.9.5, do you think? Trying to decide whether to switch to commit-by-commit to get things back working w/ Bazel vs backpin/waiting for the next release.

Neargye:
@cpsauer v0.9.5 is released

cpsauer:
Yay! Thanks :)

PR #316, opened 2023-11-15
Fix undefined behavior (`get` is moved to `magic_enum::containers`)

### LINKED ISSUE #315: Undefined behavior
Hi there!

After adding `magic_enum` to my project I got unexpected build errors on some platforms in ci. But on some platforms build was successful.

I started looking into it and found out what was going on. In `magic_enum` `std::get` is extended https://github.com/Neargye/magic_enum/blob/master/include/magic_enum/magic_enum_containers.hpp#L1130. But с++ documentation states, that "It is undefined behavior to declare a full specialization of any standard library function template." (https://en.cppreference.com/w/cpp/language/extending_std section "Function templates and member functions of templates").

I suggest to define `get` for `magic_enum` in a separate namespace. Maybe you have other suggestions?

IvanArkhipov1999:
In `magic_enum` `std::get` is extended https://github.com/Neargye/magic_enum/blob/master/include/magic_enum/magic_enum_containers.hpp#L1130. But с++ documentation states, that "It is undefined behavior to declare a full specialization of any standard library function template." (https://en.cppreference.com/w/cpp/language/extending_std section "Function templates and member functions of templates").

So, `get` is moved to `magic_enum::containers`.

Closes https://github.com/Neargye/magic_enum/issues/315

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ Header-only C++17 library provides static reflection for enums, work with any en
   color_rgb_array[Color::RED] = {255, 0, 0};
   color_rgb_array[Color::GREEN] = {0, 255, 0};
   color_rgb_array[Color::BLUE] = {0, 0, 255};
-  std::get<Color::BLUE>(color_rgb_array) // -> RGB{0, 0, 255}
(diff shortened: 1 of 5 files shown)


Neargye:
Thanks!

PR #322, opened 2023-12-06
Add support for overriding MAGIC_ENUM_ASSERT

justinboswell:
Minor change that allows overriding the implementation of `MAGIC_ENUM_ASSERT` via macro, either in `MAGIC_ENUM_CONFIG_FILE` or before including `magic_enum.hpp`


Neargye:
Thanks!

PR #323, opened 2023-12-08
checking that the enum is valid (not empty and not forward declaration)

Neargye:


include/magic_enum/magic_enum.hpp
```diff
@@ -1255,14 +1272,17 @@ template <typename E, detail::enum_subtype S = detail::subtype_v<E>>
...
 }
 
 // Obtains index in enum values from static storage enum variable.
 template <auto V, detail::enum_subtype S = detail::subtype_v<std::decay_t<decltype(V)>>>
-[[nodiscard]] constexpr auto enum_index() noexcept -> detail::enable_if_t<decltype(V), std::size_t> {
-  constexpr auto index = enum_index<std::decay_t<decltype(V)>, S>(V);
+[[nodiscard]] constexpr auto enum_index() noexcept -> detail::enable_if_t<decltype(V), std::size_t> {\
+  using D = std::decay_t<decltype(V)>;
+  constexpr auto index = enum_index<D, S>(V);
+  static_assert(detail::is_valid_enum_v<D, S>, "magic_enum requires enum implementation and valid max and min.");
```
ts826848, line 1284: Should this `static_assert` go above the `enum_index<D, S>(V)` call? The `static_assert` in that call should be run first so either that one will fire first, rendering this one moot, or neither of them will.

include/magic_enum/magic_enum_containers.hpp
```diff
@@ -331,7 +331,7 @@ template <typename E, typename V, typename Index = default_indexing<E>>
 struct array {
   static_assert(std::is_enum_v<E>);
   static_assert(std::is_trivially_constructible_v<Index>);
-  static_assert(enum_count<E>() == 0 || Index::at(enum_values<E>().front())); // check Index is constexpr
+  static_assert(enum_count<E>() > 0 && Index::at(enum_values<E>().front()));
```
ts826848, line 334: This and the similar check for `bitset` would fail even if `MAGIC_ENUM_NO_CHECK_VALID_ENUM_TYPE` is set. Is that intended?

include/magic_enum/magic_enum_fuse.hpp
```diff
@@ -90,6 +90,7 @@ template <detail::enum_subtype S, typename... Es>
 // Returns a bijective mix of several enum values. This can be used to emulate 2D switch/case statements.
 template <typename... Es>
 [[nodiscard]] constexpr auto enum_fuse(Es... values) noexcept {
+  static_assert((std::is_enum_v<std::decay_t<Es>> && ...), "magic_enum::enum_fuse requires enum type.");
```
ts826848: Is it intended that this doesn't check `count_v` as well?

include/magic_enum/magic_enum.hpp
```diff
@@ -881,6 +881,17 @@ constexpr U values_ors() noexcept {
   return ors;
 }
 
+template <typename E, enum_subtype S>
+struct is_valid_enum
+#if defined(MAGIC_ENUM_NO_CHECK_VALID_ENUM_TYPE)
```
ts826848, line 886: Would allowing validity checks to be turned on/off on a per-enum basis be worth the extra code?

thirtythreeforty: What's the reason to turn off the validity checks in the first place? Is that ever a meaningful operation?

include/magic_enum/magic_enum.hpp
Neargye: Fixed

include/magic_enum/magic_enum_containers.hpp
Neargye: fixed

Neargye: Yes, it seems logical to me to prohibit empty enums

include/magic_enum/magic_enum.hpp
Neargye: > What's the reason to turn off the validity checks in the first place? Is that ever a meaningful operation?

Maybe for backward compatibility, for example.

Neargye: > Would allowing validity checks to be turned on/off on a per-enum basis be worth the extra code?

This seems unnecessary to me.

PR #325, opened 2023-12-20
Bazel: --incompatible_use_platforms_repo_for_constraints migrated

cpsauer:
Please see https://github.com/Neargye/magic_enum/issues/309#issuecomment-1863636137


Neargye:
Thanks!

zaucy:
thanks for tackling this :)

still a mystery as to why `.bazelversion` isn't being respected.

cpsauer:
You're very welcome. Thanks again for a great library.
(Will reply re bazelversion back in #309 to keep things centralized.)

PR #329, opened 2023-12-29
CI: hash-pin sensitive workflow and configure dependabot to keep them updated

diogoteles08:
Closes #328

As mentioned on the issue, this PR enhances project security by hash-pinning the dependencies that are called under dangerous permissions. Additionally, it enables dependabot to update them automatically.

I configured dependabot in a way that all of version updates will be collapsed in a single PR sent monthly -- this avoids noisy PRs, which is a common concern haha. Regardless of the frequency chosen, for the case of security updates a PR with the fixed version would be sent right away.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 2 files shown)


Neargye:
Thanks!

PR #336, opened 2024-02-09
example_switch.cpp: properly return 0

stephanlachnit:
The main function in example_switch.cpp was declared as `int main()` but actually does not return a value - this fixes this by adding `return 0` to the end of the function.


Neargye:
Thanks!

ZXShady:
I know it is old but main automaticly returns 0 if no return statement is found

PR #343, opened 2024-04-15
Support C++20 module.

stripe2933:
A simple C++20 module export file. Importing this file using a build system will make enable user to use `import magic_enum;`. I'm also looking forward to use standard library module for faster compilation and MSVC support.


kelteseth:
Please create a PR https://github.com/kelteseth/arewemodulesyet/ if this is merged so we can track it at https://arewemodulesyet.org/

PR #350, opened 2024-05-11
Add MAGIC_ENUM_USE_STD_MODULE option to use standard library module.

stripe2933:
This PR adds two functionalities:

- When user declares `MAGIC_ENUM_USE_STD_MODULE` macro, standard library header inclusions are skipped, and it uses `import std;` instead. It prevents multiple header compilation and improves the build time. This only affects to the codebase when user use `magic_enum.cppm` module and declare the macro. Similar structure to [argparse.cppm](https://github.com/p-ranav/argparse/blob/master/module/argparse.cppm).
- Export `std::formatter` template specialization, which was missing in PR #343.

---

Fix: Commit `a3934f7`

- Move `magic_enum.cppm` module file to `module` directory, since module is not header file and doesn't match to the meaning of `include`.


Neargye:
Hi,
Could you please add minimal example?

stripe2933:
[Here](https://github.com/stripe2933/magic-enum-module-test)'s the sample project with CMake, including Clang CI.

Neargye:
Thanks!

PR #351, opened 2024-05-19
Remove redundant const from "constexpr const"

albertvaka:
According to https://en.cppreference.com/w/cpp/language/constexpr, `constexpr` implies `const`.


albertvaka:
Closing since compilers complain about `warning: ISO C++11 does not allow conversion from string literal to 'char *const' [-Wwritable-strings]`. Probably because it's `constexpr char*` and not `char constexpr*` or something like that.

PR #358, opened 2024-06-03
chore: remove crlf line endings in bazel files

zaucy:
when I originally added these bazel files I accidentally submitted some with CRLF line endings and they're causing issues on the BCR submission https://github.com/bazelbuild/bazel-central-registry/pull/1903


Neargye:
Thanks!

PR #360, opened 2024-06-08
Add better support for Yocto/PetaLinux in cmake file

ShaharHD:
Added a new option of `MAGIC_ENUM_OPT_INSTALL_PACKAGE_XML` to allow disabling the `package.xml` file when use the repo in bitbake files.

As this is a header-only library, having the `package.xml` creates an un-needed artifact on the target rootfs.

`MAGIC_ENUM_OPT_INSTALL_PACKAGE_XML` is non-breaking and using `MAGIC_ENUM_OPT_INSTALL` as its default value.


Neargye:
Thanks

ShaharHD:
> Thanks

Thank you for an awesome library. Makes so much easier to work with enums - should have been part of the C++ standard IMHO 👍 

Any ETA on next release?

PR #369, opened 2024-07-24
Update comments regarding max and min

ricardofdc:
The comments in the code regarding MAGIC_ENUM_RANGE_MIN and MAGIC_ENUM_RANGE_MAX values were not updated according to the latest changes on those default values.


Neargye:
Thanks

Issue #314, opened 2023-11-14
enum_flags limitation?


Benualdo:
Hi, I did not know about the magic_enum::enum_flags funcs until recently but the limitations are unclear to me. 

e.g. I use
```
#define MAGIC_ENUM_RANGE_MIN 0
#define MAGIC_ENUM_RANGE_MAX 256
```
and I have this enum
```
enum class Flags : u64
{
  A = 0x0000000000000001,
  B = 0x0000000000000002,
  C = 0x0000000000000004, 
  D = 0x0000000000000008, 
  E = 0x0000000000000010, 
  F = 0x0000000000000020,
  G = 0x0000000000000040, 
  H = 0x0000000000000080,  
  I = 0x0000000000000100,
  J = 0x0000000000000200,
  K = 0x0000000000000400
};
```

magic_enum::enum_name returns empty string for anything above 256 as expected but magic_enum::enum_flags_name(1026) works and returns "B|K".

Does the MAGIC_ENUM_RANGE_MAX apply to enum flags and how? Is it the maximum amount of flags avaible? Thanks

Neargye:
Hi,

`MAGIC_ENUM_RANGE_MAX` does not affect the maximum amount of flags.

enum_flags_* сhecks all `1<<(1 ... bit_count<Enum>)` // bit_count = CHAR_BIT * sizeof(Enum)
but the limitation is that it doesn't check logical AND and '0' not сhecks

Example
```cpp
enum class TestEnum {
A = 0,
B = 1,
C = 2,
D = 3,
E = 8192
};

enum_count<TestEnum>() = 4;
// '' - mean enmpty string
enum_name(TestEnum::A) = 'A'; vs enum_flags_name(TestEnum::A) = '';
enum_name(TestEnum::B) = 'B'; vs enum_flags_name(TestEnum::B) = 'B';
enum_name(TestEnum::C) = 'C'; vs enum_flags_name(TestEnum::C) = 'C';
enum_name(TestEnum::D) = 'D'; vs enum_flags_name(TestEnum::D) = 'B|C';
enum_name(TestEnum::E) = '';  vs enum_flags_name(TestEnum::E) = 'E';

enum_flags_cast('A') = nullopt; vs enum_cast('A') = TestEnum::A;
enum_flags_cast('D') = nullopt; vs enum_flags_cast('B|C') = TestEnum::D; vs  enum_cast('D') = TestEnum::D;

```

Neargye:
[...]

PR #376, opened 2024-09-04
enum_flags_name: return name for 0, if it exists

masbug:
Added constexpr that returns name for the value zero, if it exists.

### DIFF
--- a/include/magic_enum/magic_enum_flags.hpp
+++ b/include/magic_enum/magic_enum_flags.hpp
@@ -70,6 +70,14 @@ template <typename E>
   constexpr auto S = detail::enum_subtype::flags;
   static_assert(detail::is_reflected_v<D, S>, "magic_enum requires enum implementation and valid max and min.");
 
+  // There might be a name for value 0 (when no flags are set).
+  if constexpr(enum_contains<E>(0)) {
+    if(static_cast<U>(value) == 0) {
+      constexpr string_view zeroName = detail::enum_name_v<D, static_cast<D>(0)>;
+      return string(zeroName);
+    }
+  }
+
   string name;
   auto check_value = U{0};
   for (std::size_t i = 0; i < detail::count_v<D, S>; ++i) {


ZXShady:
Since this changes behavior it should probably be opt in

masbug:
I don't think it should be opt in. It fixes a bug. If 0 isn't defined as an enum value then it will still work as before.
```
// this works same as before
enum E1 {
    A = 1<<0,
    B = 2<<1
}
```
vs.
```
// this will now resolve value 0 to "None"
enum E2 {
    None = 0,
    A = 1<<0,
    B = 2<<1
}```

**Comment by ZXShady** [comment]
> I don't think it should be opt in. It fixes a bug. If 0 isn't defined as an enum value then it will still work as before.
> ```
> // this works same as before
> enum E1 {
>     A = 1<<0,
>     B = 2<<1
> }
> ```
> vs.
> ```
> // this will now resolve value 0 to "None"
> enum E2 {
>     None = 0,
>     A = 1<<0,
>     B = 2<<1
> }```


I understand, but `magic_enum::enum_values()[0]` isn't 0 either, so this seems inconsistent to only apply it here, how about `enum_cast<Bitflag>("None")` ?

**Comment by masbug** [comment]
It already works for both cases.
See this test https://godbolt.org/z/nEKoP6j4n
```c++
#include <magic_enum/magic_enum.hpp>
#include <iostream>

enum class E1 : int {
    None = 0,
    Flag1 = 1 << 0,
    Flag2 = 2 << 1
};

int main() {
    static_assert(magic_enum::enum_values<E1>()[0] == E1::None);
    static_assert(magic_enum::enum_cast<E1>("None") == E1::None);

    return 0;
}
```

ZXShady:
@masbug 

This is not a bitflag, you must do this


```cpp
template <>
struct magic_enum::customize::enum_range<E1> {
  static constexpr bool is_flags = true;
};

// now your code fails
```

otherwise you will be reflecting from `[-128,127]` instead of `N^2`

@Neargye 

It seems alot of people forget to set this boolean to true themselves, shouls it auto deduce?

masbug:
@ZXShady
Thanks for pointing that out. I see now what you mean.
I suppose this is the correct behavior for flags, since 0 is a special state.
```c++
static_assert(magic_enum::enum_values<E1>()[0] == E1::Flag1);
```
But this is missing:
```c++
static_assert(magic_enum::enum_cast<E1>("None") == E1::None);
```

ZXShady:
There was some discussion about it including the 0 value.


#314

Neargye:
@ZXShady 
> It seems alot of people forget to set this boolean to true themselves, shouls it auto deduce?

It would be great, but I haven't found a reliable way to detect that these are flags yet.


Personally, I don't think zero should be included by default. I would prefer two separate options: flags and flags_with_zero (or flags_and_zero), though this is open for discussion.

ZXShady:
> @ZXShady 
> It would be great, but I haven't found a reliable way to detect that these are flags yet.


I do it myself by checking for & and | operators if they are overloaded.

Neargye:
Could try adding auto detection. Then we definitely need swith using namespace magic_enum::bitwise_operators to enable_bitwise_operators = true 😁

PR #383, opened 2024-10-17
meson: install headers and pkg-config file

stephanlachnit:
This is useful when embedding `magic_enum` as a subproject in a larger Meson project. Would be super cool if you tag a version with this for us to use :)


Neargye:
Thanks!

PR #389, opened 2024-11-15
Only prepend $prefix for relative paths

fliiiix:
In cases where CMAKE_INSTALL_LIBDIR is already
an absolute path adding a $prefix will be incorrect.

See also issue on nix pkgs:
https://github.com/NixOS/nixpkgs/issues/144170
And current workaround:
https://github.com/NixOS/nixpkgs/blob/7eee17a8a5868ecf596bbb8c8beb527253ea8f4d/pkgs/development/libraries/magic-enum/default.nix#L21

This should be fully backwards compatible for people that want to have it relocatable.


fliiiix:
Hm seems the installation test is not happy with this change :crying_cat_face:

fliiiix:
Not me completely forgetting how bools work and what i want to do :see_no_evil:

Neargye:
Now all looks good, thanks!

PR #394, opened 2024-12-21
Support fmt and std::format in C++20 module

CrackedMatter:
~`<fmt/base.h>` defines `FMT_VERSION`, which is what magic_enum checks for when specializing `fmt::formatter`.~

https://github.com/Neargye/magic_enum/commit/ff6e5dd1c82a0c8cf8f774759933cc0a353fa5f9 removed the `<format>` and `<fmt/format.h>` includes from `magic_enum_format.hpp`.


Neargye:
Thanks!

Neargye:
Would you happen to have any suggestions regarding this problem? https://github.com/Neargye/magic_enum/issues/396

PR #406, opened 2025-04-10
Typo in README

elykwilliams:
The "Basic" example doe not work as written. 

Currently, it has `enum class Color : { RED = -10, BLUE = 0, GREEN = 10 };` which is not valid C++ syntax.


Neargye:
Thanks!

PR #407, opened 2025-04-15
fix: warnings about included file in module purview in MSVC

helynranta:
fixes warnings in MSVC about unknown pragma clang and suppresses 5244 about including file in purview of module


Neargye:
Hi @helynranta, this PR not work on linux and mac

Neargye:
Thanks!

Issue #381, opened 2024-10-13
please add an ADL version of customize::enum_range<> (code included)


lsemprini:
Hi, thanks so much for magic_enum!

As far as I can tell, the mechanism for turning on `is_flags` or other `enum_range` flags (specializing `magic_enum::customize::enum_range<T>`) only works from global scope (because C++ doesn't let us change `magic_enum::customize::` unless the scope we are in (global) encloses `magic_enum`).

This is quite inconvenient when the enum being customized is deep inside one or more classes and/or namespaces.  And it's a little dangerous too, because if we accidentally use the enum with magic_enum before the specialization, magic_enum will do the wrong thing.  It's good to have the magic_enum customize code right after where the enum is defined.

ADL is exactly the right tool for this because it can give magic_enum access to whatever scope contains the enum (even if it's deep inside namespaces and classes)

So, I have a simple proposal to add a second, ADL-based mechanism for magic_enum to find `enum_range` parameters.

The idea is for magic_enum to also check for the ability to call the function `magic_enum_enum_range(t)` and if a call is possible, use the struct returned by the function in exactly the same way that magic_enum currently uses `magic_enum::customize::enum_range<T>`  The function is never actually called at runtime (and it's constexpr) and it's never used in evaluated context, only used inside `decltype()` so it doesn't add any overhead.

Here is the code for how to do this to fetch `is_flags` (tested on MSVC19.41 with `-std:c++latest`, clang 18.0.2 with `-std=c++2c`, gcc 13.1.0 with `-std=c++20`)

Here is the existing, unmodified code from magic_enum.hpp:

```
template <typename T, typename = void>
struct has_is_flags : std::false_type {};

template <typename T>
struct has_is_flags<T, std::void_t<decltype(customize::enum_range<T>::is_flags)>> : std::bool_constant<std::is_same_v<bool, std::decay_t<decltype(customize::enum_range<T>::is_flags)>>> {};
```

Add one more struct to try the ADL case:

```
// you must define an ADL-reachable function magic_enum_enum_range(T t):
//
//    typedef enum e1 {} e1;
//    struct e1_enum_range
//    {
//        // here is the same stuff you would put into magic_enum::customize::enum_range<T>
//        inline static bool is_flags = true;
//    };
//    // include "friend" if this function is in a class
//    // exclude "friend" if this function is in a namespace (including global scope)
//    friend inline constexpr auto magic_enum_enum_range(e1) { return e1_enum_range(); }
//
template <typename T>
[...]

PR #413, opened 2025-06-09
Add ADL way of defining `enum_range`

ZXShady:
Implements #381

### DIFF
--- a/doc/limitations.md
+++ b/doc/limitations.md
@@ -23,6 +23,32 @@
 
 * If an enum is declared as a flag enum, its zero value will not be reflected.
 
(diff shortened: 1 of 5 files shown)


lsemprini (commented):
Neat!  Much cleaner than my code.  

What cases is `magic_enum::customize::adl_info<true>` meant to be used for? 

Looks like line 196 of include/magic_enum/magic_enum.hpp should say `bool` instead of `int`

Thanks!

ZXShady:
> 
> Neat!  Much cleaner than my code.  
> 
> What cases is `magic_enum::customize::adl_info<true>` meant to be used for? 
> 
> Looks like line 196 of include/magic_enum/magic_enum.hpp should say `bool` instead of `int`
> 
> Thanks!

`adl_info<true>` means `is_flags` is true while `adl_info<-10,10>` means min max range



and good catch for the `int` in `is_flags` surprised I got no warnings (likely because it is all constexpr and the conversion is valid)

lsemprini:
I see.  We should add documentation too.  I made suggested documentation changes in doc/limitations.md and doc/reference.md in my repository (that was a fork of your repository) and submitted a pull request.  I have no idea what I'm doing with GitHub, so apologies if I'm using the wrong features to do the suggestion. I can submit the files another way if you like.

https://github.com/ZXShady/magic_enum/pull/1

Neargye:
Thanks!

Issue #402, opened 2025-02-20
Feature request: remove prefix from enum name


durka:
Note: this issue is a reopening of #97.

I'm working with a lot of enums from a C library that look like this:

```
enum FOO_BAR {
    FOO_BAR_BAZ,
    FOO_BAR_QUUX,
    // etc
};
```

So when I call `enum_name()` I get `"FOO_BAR_BAZ"`, or if it's a bitflag I might get `"FOO_BAR_BAZ|FOO_BAR_QUUX"` etc. But in a lot of cases I already know I'm dealing with a `FOO_BAR` and this is noise. I would like the strings to omit `FOO_BAR_`.

Existing solutions to this problem:

1. Define another enum with the same variants and cast to that one before calling `enum_name`/`enum_flags_name`. The issue with this is that the enums are large and in third party generated code so the point of using magic enum is to reduce such chores (I'm replacing old code that is full of `const map<FOO_BAR, string>`).
2. Remove the prefix at runtime using `enum_name(v).substr(8)`. This is what I'm doing now and it works but is extra computation. It's also more cumbersome for the flags case where I have to do `replace_all(enum_flags_name(f), "FOO_BAR_", "")`.

My desired solution:

A boolean or other setting in the `customize` struct that makes magic_enum do this for me automagically.

```
template <>
struct magic_enum::customize::enum_range<FOO_BAR> {
    // one possibility, assuming the prefix is always the name of the enum plus an underscore
    constexpr bool remove_name_prefix = true;

    // more flexible but maybe harder to implement?
    constexpr char* prefix = "FOO_BAR_";

    // another idea
    constexpr int prefix_len = 8;
};
```

There could also be a flag to pass to `enum_[flags_]name` to override the setting and keep the prefix.

ZXShady:
Hi @durka 

This is a pain point when working with C-style enums. For anyone seeking an immediate solution, I’ve implemented **compile-time prefix stripping** in my  library [enchantum](https://github.com/ZXShady/enchantum) using a traits-based approach:  

```cpp  
enum FOO_BAR { FOO_BAR_BAZ, FOO_BAR_QUUX };  

template<>  
struct enchantum::enum_traits<FOO_BAR> {  
    constexpr static auto prefix_length = sizeof("FOO_BAR_") - 1; 

   constexpr static int min{ENCHANTUM_MIN_RANGE};

constexpr static int max{ENCHANTUM_MAX_RANGE};

};  

// Automatically strips prefixes:  
[...]

PR #414, opened 2025-06-12
Add prefix trimming

ZXShady:
Implements #402 Sorry @durka for the later than promised PR!

@lsemprini can you tell me what should I do with ADL info? I didn't want to add a third integral parameter I think that would be too confusing or what do you think?

Is the docs alright Daniil? I don't know how to write a decent one.

I don't have a setting to keep the prefix if turned of explicitly.

is that needed?

### DIFF
--- a/doc/limitations.md
+++ b/doc/limitations.md
@@ -23,29 +23,23 @@
 
 * If an enum is declared as a flag enum, its zero value will not be reflected.
 
-* Or, for enum types that are deeply nested in classes and/or namespaces, declare a function called `my_adl_info_struct adl_magic_enum_define_range(my_enum_type)` in the same namespace as `my_enum_type`, which magic_enum will find by ADL (because the function is in the same class/namespace as `my_enum_type`), and whose return type is a struct with `static constexpr` data members containing the same parameters as `magic_enum::customize::enum_range<my_enum_type>`
-  ```cpp
-  namespace Deeply::Nested::Namespace {
-  enum class my_enum_type { ... };
-  struct my_adl_info_struct {
-    static constexpr bool is_flags = true;
-    // you can also set min and max here (see Enum Range below)
-    // static constexpr int min = ...;
-    // static constexpr int max = ...;
-  };
-  // - magic_enum will find this function by ADL
-  // - no need to ever define this function
-  my_adl_info_struct adl_magic_enum_define_range(my_enum_type); 
-  }
-  ```
+* Or, for enum types that are deeply nested in classes and/or namespaces, declare a function called `adl_magic_enum_define_range(my_enum_type)` in the same namespace as `my_enum_type`, which magic_enum will find by ADL (because the function is in the same class/namespace as `my_enum_type`), and whose return type is a `magic_enum::customize::adl_info`.
 
-* As a shorthand, if you only want to set `is_flags` and not `min` or `max`, you can also use `magic_enum::customize::adl_info<is_flags_bool>` to avoid having to define `my_adl_info_struct` in your code:
   ```cpp
   namespace Deeply::Nested::Namespace {
-  enum class my_enum_type { ... };
+  enum class my_enum_type { my_enum_value1,my_enum_value2 };
+
   // - magic_enum will find this function by ADL
-  // - no need to ever define this function
-  magic_enum::customize::adl_info<true> adl_magic_enum_define_range(my_enum_type); 
+  // - uses builder pattern
+  // - use auto to not have to name the type yourself
+  auto adl_magic_enum_define_range(my_enum_type)
+  {
+    return magic_enum::customize::adl_info()
+    .minmax<10,10>() // the min max search range
+    .flag<true>() // whether it is a flag enum
+    .prefix<sizeof("my_enum_")-1>(); // how many characters to trim from the start of each enum entry.
+  }
+
   }
   ```
 
@@ -76,32 +70,6 @@
     static constexpr int max = 300;
     // (max - min) must be less than UINT16_MAX.
   };
-  ```
-
-* Or, for enum types that are deeply nested in classes and/or namespaces, declare a function called `my_adl_info_struct adl_magic_enum_define_range(my_enum_type)` in the same namespace as `my_enum_type`, which magic_enum will find by ADL (because the function is in the same class/namespace as `my_enum_type`), and whose return type is a struct with `static constexpr` data members containing the same parameters as `magic_enum::customize::enum_range<my_enum_type>`
-  ```cpp
-  namespace Deeply::Nested::Namespace {
-  enum class my_enum_type { ... };
-  struct my_adl_info_struct {
-    static constexpr int min = 100;
-    static constexpr int max = 300;
-    // you can also set is_flags here
-    // static constexpr bool is_flags = true;
-  };
-  // - magic_enum will find this function by ADL
-  // - no need to ever define this function
-  my_adl_info_struct adl_magic_enum_define_range(my_enum_type); 
-  }
-  ```
-
-* As a shorthand, if you only want to set `min` and `max` and not `is_flags`, you can also use `magic_enum::customize::adl_info<min_int, max_int>` to avoid having to define `my_adl_info_struct` in your code:
-  ```cpp
-  namespace Deeply::Nested::Namespace {
-  enum class my_enum_type { ... };
-  // - magic_enum will find this function by ADL
-  // - no need to ever define this function
-  magic_enum::customize::adl_info<100 /*min*/, 300 /*max*/> adl_magic_enum_define_range(my_enum_type); 
-  }
   ```
 
 ## Aliasing
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -8,6 +8,7 @@
 * [`enum_name` returns name from enum value.](#enum_name)
 * [`enum_names` obtains string enum name sequence.](#enum_names)
 * [`enum_entries` obtains pair (value enum, string enum name) sequence.](#enum_entries)
+* [`customize::enum_range`](#customizeenum_range)
 * [`enum_index` obtains index in enum value sequence from enum value.](#enum_index)
 * [`enum_contains` checks whether enum contains enumerator with such value.](#enum_contains)
 * [`enum_reflected` returns true if the enum value is in the range of values that can be reflected..](#enum_reflected)
@@ -62,6 +63,7 @@
   #define MAGIC_ENUM_RANGE_MIN 0
   #define MAGIC_ENUM_RANGE_MAX 255
   ```
+
 
 ## `enum_cast`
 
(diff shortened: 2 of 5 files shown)


lsemprini:
> Implements #402 Sorry @durka for the later than promised PR!
> 
> @lsemprini can you tell me what should I do with ADL info? I didn't want to add a third integral parameter I think that would be too confusing or what do you think?

DEFINITELY `prefix_length` needs to be settable by ADL too!  Anything that is in `enum_range` should be settable by either `customize` specialization or ADL.  Otherwise there is no way to use the feature for enums that are deep inside namespaces, and `prefix-length` is just as useful in that case.  It would be confusing and odd not to have all the `enum_range` parameters consistently available by either `customize` specialization or ADL.

As far as the `adl_info` shorthand struct, if we are going to have `adl_info`, then it should be able to provide shorthand for any of the features of `enum_range`.   The advantage of `adl_info` is you don't have to define any struct.  The disadvantage of `adl_info` is you cannot combine settings of `is_flags`, `min/max`, or `prefix_length`.  For `adl_info` may I suggest renaming them to `adl_magic_enum_range_is_flags`, `adl_magic_enum_range_min_max`, and adding `adl_magic_enum_range_prefix_length` so that it is more clear what they do when people read code, and what the parameters mean, and so that the integer of `prefix_length` no longer conflicts with the bool of `is_flags` ?  That way, we don't have to be at the mercy of having different C types.

> 
> Is the docs alright Daniil? I don't know how to write a decent one.

We would need to add more docs to limitations.md and change the existing docs for `is_flags` and `min/max` a little.  

I am happy to do that after you have finalized what the code/API should be like.

ZXShady:
> > Implements #402 Sorry @durka for the later than promised PR!
> > 
> > @lsemprini can you tell me what should I do with ADL info? I didn't want to add a third integral parameter I think that would be too confusing or what do you think?
> 
> DEFINITELY `prefix_length` needs to be settable by ADL too!  Anything that is in `enum_range` should be settable by either `customize` specialization or ADL.  Otherwise there is no way to use the feature for enums that are deep inside namespaces, and `prefix-length` is just as useful in that case.  It would be confusing and odd not to have all the `enum_range` parameters consistently available by either `customize` specialization or ADL.
> 
> As far as the `adl_info` shorthand struct, if we are going to have `adl_info`, then it should be able to provide shorthand for any of the features of `enum_range`.   The advantage of `adl_info` is you don't have to define any struct.  The disadvantage of `adl_info` is you cannot combine settings of `is_flags`, `min/max`, or `prefix_length`.  For `adl_info` may I suggest renaming them to `adl_magic_enum_range_is_flags`, `adl_magic_enum_range_min_max`, and adding `adl_magic_enum_range_prefix_length` so that it is more clear what they do when people read code, and what the parameters mean, and so that the integer of `prefix_length` no longer conflicts with the bool of `is_flags` ?  That way, we don't have to be at the mercy of having different C types.

I mean I can add a third optional parameter for adl_info

```cpp
adl_info<-10,10> // min,max
adl_info<true> // is_flags
adl_info<true,10> // is_flags + prefix length
adl_info<-10,10,10> // min,max + prefix length
```
as you noticed I don't have `adl_info<10>` to mean `prefix_length` this is intentional if you override prefix_length you should override min max anyways and why I force overriding min max if you just want to override prefix length.

I don't want to go 3 adl functions route since it will slow down compilation unnecessarily (not by much given how expensive magic enum is relative to 3 adl calls)

We can split the adl_info into 2 structs `adl_info_flags` and `adl_info_minmax` this way it is clearer than a simple boolean type to differentaite
> > 
> > Is the docs alright Daniil? I don't know how to write a decent one.
> 
> We would need to add more docs to limitations.md and change the existing docs for `is_flags` and `min/max` a little.  
> 
> I am happy to do that after you have finalized what the code/API should be like.
> 
sure.

lsemprini:
> I mean I can add a third optional parameter for adl_info...
> as you noticed I don't have `adl_info<10>` to mean `prefix_length` this is intentional if you override prefix_length you should override min max anyways and why I force overriding min max if you just want to override prefix length.

Ok now I understand the goal.  Why is it necessary to provide `min/max` if you want to provide `prefix_length`?  Is that a limitation of the implementation?

> I don't want to go 3 adl functions route since it will slow down compilation unnecessarily (not by much given how expensive magic enum is relative to 3 adl calls)

To be clear, I wasn't proposing having 3 ADL functions.  The ADL function (the function that `magic_enum` finds by ADL lookup) would still be just one function `adl_magic_enum_define_range()` and it would return 1 struct with up to all 4 params in it (`is_flags`, `min/max`, `prefix_length`).

Instead, I was just proposing renaming `adl_info` to make it more clear what the parameters mean, to someone reading code with `adl_info` in it.

> We can split the adl_info into 2 structs `adl_info_flags` and `adl_info_minmax` this way it is clearer than a simple boolean type to differentaite

How about splitting adl_info into  `adl_info_is_flags` `adl_info_minmax` and `adl_info_minmax_prefix_length`?  Then every use of the templates will document what the parameters mean to someone reading the code.

ZXShady:
> 
> > I mean I can add a third optional parameter for adl_info...
> > as you noticed I don't have `adl_info<10>` to mean `prefix_length` this is intentional if you override prefix_length you should override min max anyways and why I force overriding min max if you just want to override prefix length.

Not a limitation but if you know the prefix length then why not set the min max as well? this is what I do in my library, 

if you want the default anyway, set the `max` to `MAGIC_ENUM_RANGE_MAX`.


> Ok now I understand the goal.  Why is it necessary to provide `min/max` if you want to provide `prefix_length`?  Is that a limitation of the implementation?
> 
> > I don't want to go 3 adl functions route since it will slow down compilation unnecessarily (not by much given how expensive magic enum is relative to 3 adl calls)
> 
> To be clear, I wasn't proposing having 3 ADL functions.  The ADL function (the function that `magic_enum` finds by ADL lookup) would still be just one function `adl_magic_enum_define_range()` and it would return 1 struct with up to all 4 params in it (`is_flags`, `min/max`, `prefix_length`).
> 
> Instead, I was just proposing renaming `adl_info` to make it more clear what the parameters mean, to someone reading code with `adl_info` in it.
> 
> > We can split the adl_info into 2 structs `adl_info_flags` and `adl_info_minmax` this way it is clearer than a simple boolean type to differentaite
> 
> How about splitting adl_info into  `adl_info_is_flags` `adl_info_minmax` and `adl_info_minmax_prefix_length`?  Then every use of the templates will document what the parameters mean to someone reading the code.

I think having 2 for is_flags and minmax and prefix length being the last optional parameter.

having 4 combinations (you missed is_flags_prefix_length) seems a lot

lsemprini:
> > How about splitting adl_info into  `adl_info_is_flags` `adl_info_minmax` and `adl_info_minmax_prefix_length`?  Then every use of the templates will document what the parameters mean to someone reading the code.
> 
> I think having 2 for is_flags and minmax and prefix length being the last optional parameter.

It would work for now, but it's unintuitive when reading code that uses it, and as we add more features into the struct returned by the ADL func in the future (e.g. we're adding `prefix_length` now and will probably add more fields in the future), there will be more and more combinations.   Is there some way to rejigger `adl_info` to cleanly document itself and not have to cover all combinations?  e.g. something like `adl_info().is_flags(true).minmax(-10,20).prefix_length(4)` but that can generate the required `static constexpr` fields at compile time?

Also worth considering: maybe just delete `adl_info` entirely?  It was a cool idea but it is essentially just an optional shorthand; maybe it's less and less helpful as there are more and more params in the struct returned by the ADL func?

ZXShady:
> 
> > > How about splitting adl_info into  `adl_info_is_flags` `adl_info_minmax` and `adl_info_minmax_prefix_length`?  Then every use of the templates will document what the parameters mean to someone reading the code.
> > 
> > I think having 2 for is_flags and minmax and prefix length being the last optional parameter.
> 
> It would work for now, but it's unintuitive when reading code that uses it, and as we add more features into the struct returned by the ADL func in the future (e.g. we're adding `prefix_length` now and will probably add more fields in the future), there will be more and more combinations.   Is there some way to rejigger `adl_info` to cleanly document itself and not have to cover all combinations?  e.g. something like `adl_info().is_flags(true).minmax(-10,20).prefix_length(4)` but that can generate the required `static constexpr` fields at compile time?
> 
> Also worth considering: maybe just delete `adl_info` entirely?  It was a cool idea but it is essentially just an optional shorthand; maybe it's less and less helpful as there are more and more params in the struct returned by the ADL func?
> 

something like `magic_enum::adl_info().is_flags<true>().minmax<-100,100>().prefix_length<8>()` and would work

```cpp
auto adl_magic_enum_range(Enum)
{
    return magic_enum::adl_info().is_flags<true>().minmax<-100,100>().prefix_length<8>();
}
```

lsemprini:
That would be cool!  Is there a way to make it work and have the members be `static constexpr` ?

ZXShady:
> That would be cool!  Is there a way to make it work and have the members be `static constexpr` ?
> 
> 

```cpp
template<bool IsFlags = false,int Min = MAGIC_ENUM_RANGE_MIN,int Max = MAGIC_ENUM_RANGE_MAX,std::size_t PrefixLength = 0>
struct adl_info_holder {
    constexpr static int max = Max;
    constexpr static int min = Min;
    constexpr static bool is_flags =IsFlags;
    constexpr static std::size_t prefix_length = PrefixLength;
  
    template<int min,int max>
    static adl_info_holder<IsFlags,min,max,PrefixLength> minmax() { return {};}
     template<bool flag>
     static adl_info_holder<flag,Min,Max,PrefixLength> flag() { return {};}

     template<std::size_t PrefixLen>
     static adl_info_holder<IsFlags,Min,Max,PrefixLen> prefix() { return {};}

};

adl_info_holder<> adl_info()
{
     return {};
}

```

lsemprini:
Neat!

lsemprini:
Let me know when you've got your final code changes in and I'll do documentation

ZXShady:
> Let me know when you've got your final code changes in and I'll do documentation
> 

I am waiting for Neargye f
opinion first

Neargye:
Hi @ZXShady 
For me, it looks good!

not critical, but from my logic we can simply name `std::size_t prefix_length` - If there is no need to remove the prefix, then it will logically be zero.

ZXShady:
> Hi @ZXShady 
> For me, it looks good!
> 
> not critical, but from my logic we can simply name `std::size_t prefix_length` - If there is no need to remove the prefix, then it will logically be zero.

it is not required to be defined as in the docs

* `prefix_length` tells `magic_enum` how many characters to remove from the start of the names for all string functions. It is not required to be defined if not defined it will be assumed to be `0`

Neargye:
Is PR ready to merge?

lsemprini:
Unless I missed something, the code discussed above to allow setting
prefix_length by ADL is not yet in the tree

ZXShady:
I think it is fine to merge this then we can in a later pr revamp the adl info struct or just remove it entirely

lsemprini:
Ok but it's not just the adlinfo struct.  If I'm looking in the right
place, the low level test for prefix_length doesn't check ADL at all right
now, so there's no way to make the setting by ADL at the moment at all.  So
that would mean enums that are deep inside classes or namespaces won't be
able to take advantage of the prefix_length future like they can take
advantage of the other features like is_flags and min/max.

lsemprini:
In other words, if you compare line 176 and line 189 of

https://github.com/ZXShady/magic_enum/blob/add_prefix_length/include/magic_enum/magic_enum.hpp

the is_flags stuff checks ADL but the prefix_length stuff does not check ADL.

Is that the correct source we are considering merging, or am I looking in the wrong place?

ZXShady:
> In other words, if you compare line 176 and line 189 of
> 
> https://github.com/ZXShady/magic_enum/blob/add_prefix_length/include/magic_enum/magic_enum.hpp
> 
> the is_flags stuff checks ADL but the prefix_length stuff does not check ADL.
> 
> Is that the correct source we are considering merging, or am I looking in the wrong place?
> 

you are right indeed it ignores prefix_length adl right now.


a fix is simple adding this line

```cpp
template<class E>
constexpr auto prefix_length_or_zero<T,std::void_t<decltype(decltype(adl_magic_enum_define_range(E{}))::prefix_length)>> = std::size_t{decltype(adl_magic_enum_define_range(E{}))::prefix_length};
```

I can't currently do this as I have extremely bad internet for the next week or so

ZXShady:
I decided to remove the ability to have custom types in place of the better builder pattern `customize::adl_info()`

```cpp

auto adl_magic_enum_define_range(Enum)
{
     return magic_enum::customize::adl_info()
    .minmax<10,10>() // the min max search range
    .flag<true>() // whether it is a flag enum
    .prefix<sizeof("my_enum_")-1>(); // how many characters to trim from the start of each enum entry.
}


```

Neargye:
Hi @ZXShady 
For such a large number of parameters, the builder will be a good option. As I understand it, is it time to merge?

ZXShady:
> Hi @ZXShady 
> For such a large number of parameters, the builder will be a good option. As I understand it, is it time to merge?

I think yes it is time to merge, I don't have anything to add. is my code quality alright?

also sorry for the delays

like for example min max being always needed even if you only want to set prefix_length is that fine? Imo it is.

Neargye:
> > Hi @ZXShady
> > For such a large number of parameters, the builder will be a good option. As I understand it, is it time to merge?
> 
> I think yes it is time to merge, I don't have anything to add. is my code quality alright?
> 
> also sorry for the delays

The code itself looks fine. I’m just reconsidering the naming: since we haven’t released yet, I’m thinking of keeping the 'namespace' prefix at the beginning of the function for clarity. Naming is always tricky for me, and I’m still unsure what would work best here  `adl_magic_enum_define_range` vs `magic_enum_define_adl_range`

ZXShady:
> > > Hi @ZXShady
> > > For such a large number of parameters, the builder will be a good option. As I understand it, is it time to merge?
> > 
> > I think yes it is time to merge, I don't have anything to add. is my code quality alright?
> > 
> > also sorry for the delays
> 
> The code itself looks fine. I’m just reconsidering the naming: since we haven’t released yet, I’m thinking of keeping the 'namespace' prefix at the beginning of the function for clarity. Naming is always tricky for me, and I’m still unsure what would work best here  `adl_magic_enum_define_range` vs `magic_enum_define_adl_range`

Thanks.

or maybe `magic_enum_define_range_adl`? I agree with you that the namespace should be first.

lsemprini:
@ZXShady builder method looks good.  So it looks like all of `min/max`, `is_flags`, and `prefix_length` will be settable by ADL via the `customize::enum_range<>` specialization on line 213 of `magic_enum.cpp`, correct?

Docs look pretty good; need a few small tweaks but maybe easier if I submit a pull request for the `.md` files to Neargye later after your pull is merged.

ZXShady:
> @ZXShady builder method looks good.  So it looks like all of `min/max`, `is_flags`, and `prefix_length` will be settable by ADL via the `customize::enum_range<>` specialization on line 213 of `magic_enum.cpp`, correct?
> 
> Docs look pretty good; need a few small tweaks but maybe easier if I submit a pull request for the `.md` files to Neargye later after your pull is merged.
> 
> 

yes all of them are settable.

durka:
Thank you all for working on my feature request!

ZXShady:
I didn't get any notifications of this being merged :/

Thanks for merging my pr!

PR #422, opened 2025-08-18
enum_name function now can receive a container of enum

Paracefas:
Now you can do something like this:
`int main () {
    enum class Foo { Bar, Baz };
    std::vector<Foo> xs { Foo::Bar, Foo::Bar, Foo::Baz };
    std::println("{}", magic_enum::enum_name(xs));
}`

### DIFF
--- a/include/magic_enum/magic_enum.hpp
+++ b/include/magic_enum/magic_enum.hpp
@@ -1363,6 +1363,15 @@ template <detail::enum_subtype S, typename E>
   return enum_name<D, S>(value);
 }
 
+// Returns a vector of names from containers' enum.
+// If one enum value of the container does not have a name or value out of range, returns empty string.
+template <typename E, template<typename, typename...> typename Container, typename... Args> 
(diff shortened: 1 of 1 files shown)


ZXShady:
Hi, I think this is out of scope of `magic_enum` magic_enum is a 0 cost library and embedding allocations inside it is breaking that.

Also why only `enum_name` how about `enum_cast` ? `enum_contains` etc...

Also the return type is always vector and the template template parameter is quite limitting to what containers can be used you cannot use `std::array` with it for example

unterumarmung:
Instead of returning `std::vector` it would be logical to return the same container type. The problem with this appoach, though, that not all the containers have `push_back` method. So it should have some kind of detection what member function to use - it is easy for C++20 with `requires` and not so easy with C++17 - but there is `is_detected` pattern that can help.

Neargye:
Unfortunately, I don't think this is the best addition to the API.
I think it's better to implement such functions locally if the project needs them. Or expand all APIs to containers, but this doesn't seem reasonable.

PR #423, opened 2025-09-01
Prevent multiple definitions of adl_info

Antidote:
As currently implemented, `adl_info` can potentially be defined wherever `magic_enum.hpp` is included, this can lead to linker errors downstream.

Making the function static-inline prevents this from happening.

### DIFF
--- a/include/magic_enum/magic_enum.hpp
+++ b/include/magic_enum/magic_enum.hpp
@@ -203,7 +203,7 @@ struct adl_info_holder {
     constexpr static adl_info_holder<IsFlags,Min,Max,prefix_len> prefix() { return {};}
 };
 
-adl_info_holder<> adl_info()
+static inline adl_info_holder<> adl_info()
 {
      return {};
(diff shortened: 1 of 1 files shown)


ZXShady:
just mark it "inline". also this should be constexpr only (no inline keyword needed)

include/magic_enum/magic_enum.hpp
```diff
@@ -203,7 +203,7 @@ struct adl_info_holder {
     constexpr static adl_info_holder<IsFlags,Min,Max,prefix_len> prefix() { return {};}
 };
 
-adl_info_holder<> adl_info()
+static inline adl_info_holder<> adl_info()
```
ZXShady, line 206:
```suggestion
constexpr adl_info_holder<> adl_info()
```

ZXShady (changes requested):

Neargye:
Thanks, fixed

PR #424, opened 2025-09-01
Fix link error

MaxwellGengYF:
Link error with duplicated symbol without "inline"


Antidote:
This is a duplicate of #423

Neargye:
Thanks, fixed

PR #430, opened 2025-10-02
Replace test option with not building tests by default

frankier:
Following a discussion [here](https://github.com/mesonbuild/wrapdb/pull/2427), it seems like it would be perhaps more standard to not build tests by default rather than using a feature flag. This way they are not built by default by projects using this as a subproject, but will still be build automatically when running `meson test` in this projects and projects depending upon it. What do you think?


Neargye:
Thanks!

PR #433, opened 2025-10-15
Fix MSVC C++20/23 module compatibility for static_str conversion operator

banderzhm:
- Add explicit str() method to static_str for explicit string_view conversion
- Conditionally disable implicit operator string_view() in MSVC module mode
- Non-module builds and other compilers remain unaffected

### DIFF
--- a/include/magic_enum/magic_enum.hpp
+++ b/include/magic_enum/magic_enum.hpp
@@ -318,7 +318,14 @@ class static_str {
 
   constexpr std::uint16_t size() const noexcept { return N; }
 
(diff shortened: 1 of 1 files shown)


ZXShady:
is static_str a public type? It shouldn't be and I think we should drop the conversion operator

banderzhm:
Clang and MSVC behave completely differently.This is the best solution I could think of.

Neargye:
Thanks for PR!
And fix in master, operator string_view() not need, cuz static_str is provate class

PR #438, opened 2025-12-06
add rules_cc bazel dependency

zaucy:
In bazel 9 all cc_* rules must be loaded from `rules_cc` or else there will be a build error.


kgreenek:
+1 for this! This is preventing the BCR CI from being able to compile anything that depends on @magic_enum using rolling.

Neargye:
Thanks!

PR #450, opened 2026-03-16
Added efficient iterators to magic_enum::containers::bitset

KenanMillet:
This change introduces iterators to magic_enum::containers::bitset.
The iterators utilize std::countr_zero (or compiler-based equivalents if below c++20) to efficiently determine what the next set bit is in the bitset, allowing you to iterate over all set Enum values within a bitset much faster than:
- a magic_enum::containers::set (especially as the number of enums added to the bitset/set increases)
- and much faster than checking each bit in the bitset individually (especially as the size of the bitset increases)

The change comes with an example, but I did not see anywhere to put benchmark tests, so I left those out.

### DIFF
--- a/example/example_containers_bitset.cpp
+++ b/example/example_containers_bitset.cpp
@@ -51,6 +51,11 @@ int main() {
   std::cout << color_bitset.test(Color::RED) << std::endl; // false
   std::cout << color_bitset.test(Color::GREEN) << std::endl; // true
   std::cout << color_bitset.test(Color::BLUE) << std::endl; // true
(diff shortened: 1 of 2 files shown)


Neargye:
@KenanMillet Could you please check build failed?

KenanMillet:
@Neargye Made some tweaks. Addressed the clang/macos problems. Turns out they have a <intrin.h> file that is just a stub. Also fixed the windows 32 issues. Lastly, fixed some bugs that I identified after looking back at the code with fresh eyes.

Neargye:
Thanks, looks good!

PR #456, opened 2026-05-13
fix / added case_insensitive to module file

Stevan1996:
When using magic enum via the C++20 module, `magic_enum::case_insensitive` cannot be resolved because it is missing in the module file. There might be more symbols that are missing. I did not check throughoutly.


Neargye:
Thanks!

Issue #462, opened 2026-07-17
`max_v` should be moved outside of `detail`


Ambeco:
`max_v` is very useful, but seems "private" inside of the `detail` namespace.  IMO, it should be moved out.
Also, it's `enum_subtype S` template parameter should default to `= subtype_v<E>`.

It is less clear if `min_v` is equally value, but for symmetry, should probably also be made "public".  Alternatively, expose a `range_v<E>` that returns the range `min_v` to `max_v`.  

These are handy for helper code that map enum values to other things,  such as in https://stackoverflow.com/a/79981587/845092

Neargye:
Hi @Ambeco,
Instead of exposing the internal `min_v` and `max_v`, what about adding public functions that return the max/min enum values?
```cpp
constexpr auto min = magic_enum::enum_min<MyEnum>();
constexpr auto max = magic_enum::enum_max<MyEnum>();

// impl
template <typename E, detail::enum_subtype S = detail::subtype_v<E>>
[[nodiscard]] constexpr auto enum_max() noexcept -> detail::enable_if_t<E, std::decay_t<E>> {
  using D = std::decay_t<E>;
  static_assert(detail::count_v<D, S> > 0, "magic_enum::enum_max requires at least one reflected enum value.");

  return detail::values_v<D, S>.back();
}
```
They would return MyEnum, use the default enum subtype, and also support an explicit subtype:
```cpp
magic_enum::enum_max<MyEnum, magic_enum::as_flags<>>();
```

The underlying value would still be available through `enum_underlying(max`). This keeps the internal variables private and follows the existing function-based API.

PR #464, opened 2026-08-02
add enum_min and enum_max

aouxwoux:
## what changed

- add enum_min and enum_max to the public api
- support the default reflected subtype and explicit s_flags<>
- add common/flags tests and reference docs

## why

detail::min_v and detail::max_v are useful when mapping reflected enum values, but they are internal. this exposes the same information through the existing function-based api while keeping those implementation details private.

the functions return the first and last reflected values, so custom ranges and flags use the same ordering as enum_values.

addresses #462

## checks

- git diff --check
- not run locally: this environment does not have a c++ compiler/toolchain

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -4,6 +4,7 @@
 * [`enum_value` returns enum value at specified index.](#enum_value)
 * [`enum_values` returns enum value sequence.](#enum_values)
 * [`enum_count` returns number of enum values.](#enum_count)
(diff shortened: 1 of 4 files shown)


Neargye:
Please export enum_min and enum_max from module/magic_enum.cppm and add module coverage. Currently, import magic_enum; users cannot access the new public API.
And need rebase onto the current master; the branch is 19 commits behind and currently has merge conflicts.

Issue #467, opened 2026-08-09
g++-16 on OSX: test_flags-cpp26 Exception: SegFault


ClausKlein:
```bash
bash-5.3$ git describe --tag --dirty
v0.9.8-54-g00d372c
bash-5.3$ git branch
  develop
  feature/format-cmake-files
  feature/module_target
* master
  module_target
bash-5.3$ 

cmake -S . -B build/release -G Ninja -D CMAKE_BUILD_TYPE=Release -D CMAKE_CXX_STANDARD=26
ninja -C build/release
ninja -C build/release test

bash-5.3$ gmake env | egrep '^C' 
CMAKE_CONFIG_TYPE = Release
CMAKE_PREFIX_PATH = /Users/clausklein/.local
CMAKE_GENERATOR = Ninja
CURDIR = /Users/clausklein/Workspace/cpp/magic_enum
CMAKE_INSTALL_PREFIX = /Users/clausklein/.local
CMAKE_CONFIGURATION_TYPES = Release;Debug
CPM_USE_LOCAL_PACKAGES = NO
CLASSPATH = /usr/local/opt/antlr/antlr-4.13.0-complete.jar:
CMAKE_EXPORT_COMPILE_COMMANDS = YES
CMAKE_CXX_COMPILER_LAUNCHER = ccache
CMAKE_C_COMPILER_LAUNCHER = ccache
CPM_SOURCE_CACHE = /Users/clausklein/.cache/CPM
CXX = g++-16
bash-5.3$ ctest --test-dir build/release --rerun-failed --output-on-failure
Test project /Users/clausklein/Workspace/cpp/magic_enum/build/release
    Start 18: test_flags-cpp26
1/1 Test #18: test_flags-cpp26 .................***Exception: SegFault  0.01 sec
[doctest] doctest version is "2.5.2"
[doctest] run with "--help" for options
===============================================================================
/Users/clausklein/Workspace/cpp/magic_enum/test/test_flags.cpp:123:
TEST CASE:  enum_cast
  string

/Users/clausklein/Workspace/cpp/magic_enum/test/test_flags.cpp:123: FATAL ERROR: test case CRASHED: SIGSEGV - Segmentation violation signal

===============================================================================
/Users/clausklein/Workspace/cpp/magic_enum/test/test_flags.cpp:123:
TEST CASE:  enum_cast

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  string

===============================================================================
[...]

Issue #200, opened 2022-07-20
nested magic_enum::enum_switch -- failed on msvc?


Korsar13:
https://gist.github.com/Korsar13/5c171d0cd0b47d62276cac07b6b6a184

compiled by clang, prints 
```
convert from char to char
convert from char to int
convert from char to float
convert from int to char
convert from int to int
convert from int to float
convert from float to char
convert from float to int
convert from float to float
```

msvc compilation fails,
`mm.cpp(39): error C2893: Failed to specialize function template 'void convert_tmpl(const TSrc *,TDst *,size_t)'`

Korsar13:
initial code (compiled by msvc _without errors_, but **wrong result**) -- https://gist.github.com/Korsar13/f993beaa52947a0ce42a632602e2aecf

schaumb:
Hi!

The problem here is not the magic_enum library, but that MSVC handles the nested lambdas.
As you [can see](https://godbolt.org/z/qWKvzMcWT), if the nested `enum_switch` not use the outside type, the code compiles.

Probably the compiler wants to simplify lambdas instantiation, that's why It don't want to create new lambda types for all enum / different usings. 

This problem can be eliminated if you use not nested lambda, but some struct.
https://godbolt.org/z/4aYezTzvf

schaumb:
From the library side:

This issue can be eliminated with the following techniques:
- make `magic_enum::enum_for_each` and `magic_enum::enum_switch` to handle variadic arguments, and generate all enum combination calls, or the multiple switch call. The enum_switch implementation will be tricky.
These calls can be slow at compile time.
- create [magic_enum::bind_front](https://en.cppreference.com/w/cpp/utility/functional/bind_front) constexpr function for backward compatible way to c++17

Neargye:
@schaumb 
I added a new header `magic_enum_utility.hpp` for different secondary functions, maybe add `magic_enum::bind_front` there?
and add some example and note in doc

`magic_enum::bind_front` seems like a cheaper option to me.

PR #471, opened 2026-08-12
feat(utility): add bind_front for enum_switch partial application (#200)

ezralicodes:
## Summary

Adds `magic_enum::bind_front` to `magic_enum_utility.hpp` as a C++17-friendly partial-application helper. Maintainer suggested this for nested `enum_switch` use cases on MSVC ([#200](https://github.com/Neargye/magic_enum/issues/200)).

Also documents the GCC 16 C++26 `enum_flags_cast` workaround ([#467](https://github.com/Neargye/magic_enum/issues/467)).

## Test plan

- [x] New doctest `bind_front` case in `test/test.cpp`
- [ ] CI matrix (existing tests)

### DIFF
--- a/doc/limitations.md
+++ b/doc/limitations.md
@@ -21,6 +21,8 @@
 
 * Zero is not reflected for flag enums.
 
+* GCC 16 with `-std=c++26` may miscompile string parsing that uses `find`/`remove_prefix` on flag names. `enum_flags_cast` uses an explicit token loop instead; see [#467](https://github.com/Neargye/magic_enum/issues/467).
+
 ## Enum Range
 
 * For non-flag enums, range-based reflection only considers values in `[MAGIC_ENUM_RANGE_MIN, MAGIC_ENUM_RANGE_MAX]`.
--- a/include/magic_enum/magic_enum_utility.hpp
+++ b/include/magic_enum/magic_enum_utility.hpp
@@ -36,6 +36,7 @@
 
 #ifndef MAGIC_ENUM_USE_STD_MODULE
 #  include <tuple>
+#  include <utility>
 #endif
 
 namespace magic_enum {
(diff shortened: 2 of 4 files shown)


doc/limitations.md
```diff
@@ -21,6 +21,8 @@
 
 * Zero is not reflected for flag enums.
 
+* GCC 16 with `-std=c++26` may miscompile string parsing that uses `find`/`remove_prefix` on flag names. `enum_flags_cast` uses an explicit token loop instead; see [#467](https://github.com/Neargye/magic_enum/issues/467).
```
Neargye, line 24: This was never required and should be removed; internal documentation remains in the comments.

test/test.cpp
```diff
@@ -1315,6 +1315,19 @@ TEST_CASE("enum_for_each") {
   }
 }
 
+TEST_CASE("bind_front") {
```
Neargye, line 1318: The test does not verify fix for issue #200.

Neargye:
Please also export `bind_front` from `magic_enum.cppm` and cover it in `test_module.cpp`.

include/magic_enum/magic_enum_utility.hpp
```diff
@@ -144,6 +145,32 @@ template <typename E, detail::enum_subtype S = detail::subtype_v<E>>
...
+ public:
+  template <typename Fn, typename... Args>
+  constexpr explicit bind_front_t(Fn&& fn, Args&&... args)
+      : fn_(std::forward<Fn>(fn)), bound_(std::forward<Args>(args)...) {}
+
+  template <typename... CallArgs>
+  constexpr decltype(auto) operator()(CallArgs&&... call_args) const {
+    return std::apply(
+        fn_,
+        std::tuple_cat(bound_,
```
Neargye, line 159: Please avoid `tuple_cat` here: it copies bound arguments into a temporary and breaks mutable callables, lvalue references, move-only values, and reference returns. Invoke the stored arguments directly via `std::get`, preserving the wrapper value category with `&`, `const&`, `&&`, and `const&&` overloads.

PR #472, opened 2026-08-12
Install legacy magic_enum.hpp compatibility shim

### LINKED ISSUE #419: The binary installation and source code installation paths are different
Hello,

When I built package `libhri`, it needs the `magic_enum` package. So I built and installed the package from source.

For the `face.cpp` of `libhri` package, it includes `magic_enum` by this way `#include "magic_enum.hpp"`.

But it built faild with follow error:
```
/home/qiguanxiao/ws/ws_libhri/hri/src/hri/face.cpp:27:10: fatal error: magic_enum.hpp: No such file or directory
   27 | #include "magic_enum.hpp"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [CMakeFiles/hri.dir/build.make:90: CMakeFiles/hri.dir/src/hri/face.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:174: CMakeFiles/hri.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
---
Failed   <<< hri [17.3s, exited with code 2]

```

I found that the source installation paths is `ws/install/magic_enum/include/magic_enum/magic_enum.hpp`. 
So I replaced  `#include "magic_enum.hpp"` with  `#include "magic_enum/magic_enum.hpp"` in `face.cpp`. Finally it can be built correctly.

Then I used `sudo apt install ros-humble-magic-enum` to install the binary packge. 
I found that the install path is `/opt/ros/humble/include/magic_enum.hpp`.
 It does not have the prefix module.  So the way `#include "magic_enum.hpp"` is correct.
But for other packages that I used apt to install, their install path are "/opt/ros/humble/include/package/file.hpp".

So I wonder to know if I went wrong or the project itself does have problems.

Thanks!

ezralicodes:
Fixes #419

## Summary

- Install a compatibility shim header at `include/magic_enum.hpp` that forwards to `magic_enum/magic_enum.hpp`.
- Document the legacy include path for users upgrading from pre-0.9.7 layouts.

## Test plan

- [ ] `cmake --install` and verify both `include/magic_enum/magic_enum.hpp` and `include/magic_enum.hpp` exist.
- [ ] Compile a consumer with `#include <magic_enum.hpp>` against the installed package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,12 @@ if(MAGIC_ENUM_OPT_INSTALL)
     )
   endif()
 
+  # Compatibility shim for `#include <magic_enum.hpp>` (pre-0.9.7 layout).
+  install(
+    FILES "${PROJECT_SOURCE_DIR}/include/magic_enum.hpp"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+  )
+
   set(MAGIC_ENUM_EXPORT_TARGETS "${PROJECT_NAME}")
   if(MAGIC_ENUM_USE_MODULES)
     install(
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ Header-only C++17 library provides static reflection for enums, work with any en
   }
   ```
 
+  Installed packages also provide a compatibility shim so older code can keep using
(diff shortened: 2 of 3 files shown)


Neargye:
That will only make things worse; for example, package managers haven't been taken into account here.

PR #473, opened 2026-08-14
feat: bitshift operators in bitwise_operators (#300)

### LINKED ISSUE #300: Suggestion: bitshift operators
I would love to bitshift an integral value by an enum value.

Example usage:
Creating a flag enum based on an existing enum:
```
enum class A {
    X = 0,
    Y = 1,
    Z = 2
};

enum class AFlags {
    XF = 1 << A::X,
    YF = 1 << A::Y,
    ZF = 1 << A::Z
};
```
General Bitshifting by an enum value:
```
A a = A::Y;
auto flag = 1 << a;
```

Currently i have that implemented by the C++23 code:
```
template<typename E>
requires std::is_enum_v<E>
constexpr auto operator<<(const std::integral auto& a, const E& b) {
    return a << std::to_underlying(b);
}
```

I suggest implementing a bitshift in both directions by an enum. Maybe you want to add bitshifting an enum by an integral as well, and maybe even an enum by an enum.

ezralicodes:
## Summary
Adds opt-in bit-shift operators for enums in `magic_enum::bitwise_operators` (#300):

- `integral << enum` / `>>`
- `enum << integral` / `>>`
- `enum << enum` / `>>`
- matching compound assignment forms

Same opt-in style as existing `|` `&` `^` `~` operators.

## Test plan
- [ ] `1 << SomeEnum::X` works with `using namespace magic_enum::bitwise_operators`
- [ ] Existing bitwise flag tests still pass

Closes #300


Neargye:
No PR is required; please read the discussion in the issues.

PR #474, opened 2026-08-14
feat: public enum_min()/enum_max() helpers (#462)

### LINKED ISSUE #462: `max_v` should be moved outside of `detail`

ezralicodes:
## Summary
Adds public `magic_enum::enum_min()` / `enum_max()` helpers (#462) so callers do not need `detail::min_v` / `detail::max_v`.

## Test plan
- [ ] `enum_min<Color>()` / `enum_max<Color>()` return expected enumerators
- [ ] Existing tests still pass

Closes #462

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -4,6 +4,7 @@
 * [`enum_value` returns enum value at specified index.](#enum_value)
 * [`enum_values` returns enum value sequence.](#enum_values)
 * [`enum_count` returns number of enum values.](#enum_count)
+* [`enum_min` and `enum_max` return min/max reflected enum values.](#enum_min-and-enum_max)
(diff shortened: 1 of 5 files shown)


Neargye:
duplicate https://github.com/Neargye/magic_enum/pull/464

